PR #1, opened 2018-10-27
WIP: Add keyby iterator

segeljakt:
This PR adds support for a `keyby` iterator which can be used as:

```
let s = streamappender[i32];
...
let s = result(s);
for(keyby(s, |e| u64(e)), appender[i32], |b,k,v| merge(b,v));
```

`|e| u64(e)` is a function which takes an element and returns a corresponding key, and is used for partitioning elements. For example, `[1,2,3,3]` would in this case be partitioned into: `[1]`, `[2]` and `[3,3]`. The body of the for loop then iterates over each key-value, where `k` is the key and `v` is the value.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 *.class
 *.log
 /bin/
+*.tokens
 
 # sbt specific
 .cache
@@ -27,6 +28,7 @@ bin/
 .project
 .settings
 .classpath
+.idea/*
 
 etarget/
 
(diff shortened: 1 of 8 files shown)


segeljakt:
I can squash before the merge

segeljakt:
Ok, good points. I'll add some more commits

segeljakt:
I think it is best if I continue with `hash` in a separate PR #3.

PR #18, opened 2019-11-12
Fixes for master

frej:
Various things found while trying to get the frontend to be usable as a filter. Check the individual commits for the details.


segeljakt:
👍 💯

PR #23, opened 2019-11-28
Add unsigned literals, and update suffixes

segeljakt:
Do not merge, need to fix tests first


segeljakt:
Merging should now be OK.

src/main/scala/se/kth/cda/arc/ast/AST.scala
```diff
@@ -14,6 +14,7 @@ sealed trait ASTNode {
 
 object AST {
 
+
```
frej: unrelated white-space change?

segeljakt: oops

segeljakt: it is now undone 😄

segeljakt:
Tests are passing now, @frej could you take a look if there is anything to change?

.gitignore
```diff
@@ -133,3 +133,6 @@ arc-mlir/.guessed-arc-mlir
 
 # Editor backup files
 *~
+build/
+CMakeCache.txt
+CMakeFiles/
```
frej: The only reason for `CMakeCache.txt` and `CMakeFiles/` to occur in your source directory is if you did something wrong, and that could break proper builds in `build/`. So I think it is better to only add `build/`, that way we'll notice that something is wrong.

segeljakt:
Oops, pushed too many commits

segeljakt:
@frej I split it into 3 commits:
* One that fixes the Lexer tests
* One that changes the postfixes and adds tests for them.
* One that adds unsigned literals and adds tests for them.
Does it work or do you want the tests in separate commits?

frej:
Now the split of the work into logical commits which should pass their tests looks good!

Unfortunately c812f80b5a484905557e874b1416bc4eed8b0857 and 94db47868498e1a285465d3c4abf474161113f90 fails `sbt test` and because of that I cannot run `sbt asssemble` and check the mlir tests.

frej:
How do we get travis to run the tests for all commits and not just the tip?

segeljakt:
@frej :shipit:?

PR #24, opened 2019-12-03
arc-mlir: Add support for vector literals

frej:
Vectors are represented by tensors. We introduce a new variadic
operation, "arc.make_vector", for creating vectors from basic
values. A later transform pass is intended to convert constant vectors
to MLIR `dense<[...]]> : tensor<<size>x<type>>` notation.

### DIFF
--- a/src/main/scala/se/kth/cda/arc/ast/printer/MLIRPrinter.scala
+++ b/src/main/scala/se/kth/cda/arc/ast/printer/MLIRPrinter.scala
@@ -88,6 +88,17 @@ object MLIRPrinter {
       tmp
     }
 
+    def dumpVec(identifiers: SymbolMap, elems: Vector[Expr], ty: Type): String = {
+      val Vec(elemTy) = ty
+      val es = elems.map(_.toMLIR(identifiers))
+      val elementTypes = es.map(_ => elemTy.toMLIR).mkString(", ")
+      val tmp = newTmp
+      out.print(s"""${tmp} = "arc.make_vector"(${es
+        .mkString(", ")}) : (${elementTypes}) -> tensor<${es.length}x${elemTy.toMLIR}>\n""");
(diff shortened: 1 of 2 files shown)


src/main/scala/se/kth/cda/arc/ast/printer/MLIRPrinter.scala
```diff
@@ -88,6 +88,20 @@ object MLIRPrinter {
       tmp
     }
 
+    def dumpVec(identifiers: SymbolMap, elems: Vector[Expr], ty: Type): String = {
+      val elemTy = ty match {
+        case Vec(et) => et
+        case x       => x
+      }
+      val es = elems.map(e => e.toMLIR(identifiers))
```
segeljakt:
```scala
elems.map(e => e.toMLIR(identifiers))
```
can be written as:
```scala
elems.map(_.toMLIR(identifiers))
```
It's not exactly important, just stylistic 😅

src/main/scala/se/kth/cda/arc/ast/printer/MLIRPrinter.scala
```diff
@@ -88,6 +88,20 @@ object MLIRPrinter {
       tmp
     }
 
+    def dumpVec(identifiers: SymbolMap, elems: Vector[Expr], ty: Type): String = {
+      val elemTy = ty match {
```
segeljakt: Not sure but I think this can be written as:
```scala
val Vec(elemTy) = ty
```

src/main/scala/se/kth/cda/arc/ast/printer/MLIRPrinter.scala
frej: This is the kind of tips I appreciate as I can't say that I really know Scala.

PR #25, opened 2019-12-04
Add MLIR support for the Arc dialect

frej:


frej:
@segeljakt This is refreshed now, please take a look.

PR #35, opened 2019-12-16
Constant fold arc vectors

frej:


segeljakt:
Looks good to me. When we add streaming operators later on in Arc, it might be good to distinguish `"arc.sink"` from streaming sinks, but it is not necessary at the moment I think.

frej:
What about `arc.keep` instead?

segeljakt:
That should work well!

PR #36, opened 2019-12-18
arc-mlir: Remove no longer pertinent comment

frej:
We have special code in MLIRPrinter to print negated literals as just
a negative MLIR constant.


segeljakt:
👍

PR #38, opened 2019-12-20
MakeTuple operator

segeljakt:


arc-mlir/src/include/arc/arc-ops.td
```diff
@@ -193,13 +193,24 @@ def MakeTuple
...
+    unsigned NumOperands = Operation->getNumOperands();
+    auto TupleTy = Operation->getResult(0).getType().dyn_cast<TupleType>();
+    auto ElemTys = TupleTy.getTypes();
     if (NumOperands != TupleTy.size())
       return emitOpError("result does not match the number of operands: found ")
              << NumOperands << " but expected " << TupleTy.size()
              << " operands";
+    auto a = ElemTys.begin();
+    auto b = OperandTys.begin();
+    while (a != ElemTys.end() and b != OperandTys.end()) {
```
frej: C++ is strange, how did `and` go through the compiler?

I'd fix the compilation error by doing (this is untested):

```C++
    auto Operation = this->getOperation();
    unsigned NumOperands = Operation->getNumOperands();
    TupleType TupleTy = Operation->getResult(0).getType().cast<TupleType>();
    auto ElemTys = TupleTy.getTypes();
    if (NumOperands != TupleTy.size())
      return emitOpError("result does not match the number of operands: found ")
             << NumOperands << " but expected " << TupleTy.size()
             << " operands";
    unsigned I = 0;
    for (const Type &ElemTy : Operation->getOperands().getTypes()) {
        if (ElemTys[I] != ElemTy)
           return emitOpError("Operand types do not match, found ")
              << ElemTy << " but expected " << ElemTys[I];
        I++;
    }
    return mlir::success();

```

arc-mlir/src/include/arc/arc-ops.td
```diff
@@ -193,13 +193,24 @@ def MakeTuple
   let results = (outs AnyTuple);
 
   let verifier = [{
-    unsigned NumOperands = this->getOperation()->getNumOperands();
-    auto TupleTy =
-        this->getOperation()->getResult(0)->getType().dyn_cast<TupleType>();
+    auto Operation = this->getOperation();
+    auto OperandTys = Operation->getOperands().getTypes();
+    unsigned NumOperands = Operation->getNumOperands();
+    auto TupleTy = Operation->getResult(0).getType().dyn_cast<TupleType>();
```
frej: You want a `cast<>()` here, as if the tuple is not a tuple type you want the assert in debug mode. If it triggers something is very broken elsewhere.

segeljakt:
@frej I added your changes and now it seems to work! Do you think I should add a tuple indexing operator in this PR or a in new one?

segeljakt:
I forgot, I also need a test like you said for empty tuples

frej:
Looks good with the extra check too.

PR #41, opened 2020-01-07
arc-mlir: Adapt to MLIR in the LLVM repo

frej:
MLIR now lives in the LLVM repo. To adapt we:

* Remove the no longer used tensorflow-mlir sub-repo.

* Update our internal paths to account for the new location of MLIR.

* Adapt to upstream changes:

  + The range iterator for Operation::operand_range gives us const
    references.

  + Some standard operators have been renamed: divis -> divi_signed,
    diviu -> divi_unsigned, remis -> remi_signed, remiu ->
    remi_unsigned.

  + MLIR library renames: use the same libraries as mlir-opt.


segeljakt:
Looks great!

PR #48, opened 2020-01-28
Add Appender Type

segeljakt:
This adds an appender type. It works, but might be doable through table gen only. Otherwise, it still needs some diagnostics and refactoring.

### DIFF
--- a/arc-mlir/src/arc-dialect.cpp
+++ b/arc-mlir/src/arc-dialect.cpp
@@ -21,6 +21,7 @@
 //
 //===----------------------------------------------------------------------===//
 
+#include "mlir/IR/DialectImplementation.h"
 #include "mlir/IR/StandardTypes.h"
 #include "llvm/Support/raw_ostream.h"
 
@@ -28,6 +29,7 @@
 
 using namespace mlir;
 using namespace arc;
+using namespace types;
 
 //===----------------------------------------------------------------------===//
 // ArcDialect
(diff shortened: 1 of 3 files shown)


arc-mlir/src/include/arc/arc-dialect.h
```diff
@@ -23,21 +23,94 @@
...
+    //           if (loc)
+    //             context->emitError(loc) << "non integer-type passed to
+    //             'AppenderType'";
+    //           return failure();
+    //         }
+    return success();
+  }
+
+  Type getMergeType() { return getImpl()->mergeType; }
+};
```
segeljakt: This should probably be moved into a `arc-dialect.cpp` but I'm not sure which parts to transfer. How does C++ handle forward declarations compared to C?

arc-mlir/src/include/arc/arc-dialect.h
```diff
@@ -23,21 +23,94 @@
...
+    return Base::getChecked(location, context, ArcTypes::Appender, mergeType);
+  }
+
+  static LogicalResult
+  verifyConstructionInvariants(llvm::Optional<Location> loc,
+                               MLIRContext *context, Type type) {
+    //         if (!type.isa<IntegerType>()) {
+    //           if (loc)
+    //             context->emitError(loc) << "non integer-type passed to
+    //             'AppenderType'";
```
segeljakt: This code (from the [tutorial](https://mlir.llvm.org/docs/DefiningAttributesAndTypes/)) does not compile. Maybe we could remove the function as it seems to be optional ([Toy](https://mlir.llvm.org/docs/Tutorials/Toy/Ch-7/) does not implement it).

arc-mlir/src/include/arc/arc-dialect.h
```diff
@@ -23,21 +23,94 @@
 #ifndef ARC_DIALECT_H_
 #define ARC_DIALECT_H_
 
+#include "mlir/IR/StandardTypes.h"
+#include <mlir/IR/Builders.h>
 #include <mlir/IR/Dialect.h>
 #include <mlir/IR/Function.h>
-#include <mlir/IR/Builders.h>
 #include <mlir/IR/Operation.h>
```
segeljakt: @frej `clang-format` did this, should I revert it or does it matter?

frej: I have probably missed to run clang-format on that one, leave that as a separate patch as its unrelated to appenders

arc-mlir/src/include/arc/arc-dialect.h
frej:
```patch
diff --git a/arc-mlir/src/arc-dialect.cpp b/arc-mlir/src/arc-dialect.cpp
index cd8f967..bef6331 100644
--- a/arc-mlir/src/arc-dialect.cpp
+++ b/arc-mlir/src/arc-dialect.cpp
@@ -110,6 +110,8 @@ Type ArcDialect::parseType(DialectAsmParser &parser) const {
   return nullptr;
 }
 
+Type ArcTypes::AppenderType::getMergeType() { return getImpl()->mergeType; };
+
 //===----------------------------------------------------------------------===//
 // Arc Type Printer
 //===----------------------------------------------------------------------===//
diff --git a/arc-mlir/src/include/arc/arc-dialect.h b/arc-mlir/src/include/arc/arc-dialect.h
index d90ffec..b98395c 100644
--- a/arc-mlir/src/include/arc/arc-dialect.h
+++ b/arc-mlir/src/include/arc/arc-dialect.h
@@ -95,7 +95,7 @@ public:
     return success();
   }
 
-  Type getMergeType() { return getImpl()->mergeType; }
+  Type getMergeType();
 };
 } // namespace ArcTypes
```
is how to move an inline declaration to the module.

arc-mlir/src/include/arc/arc-dialect.h
segeljakt: Edit: I talked to the MLIR devs, apparently it is deprecated. Instead, we can use `mlir::emitOptionalError`.

arc-mlir/src/include/arc/arc-dialect.h
segeljakt: Oh, then it makes sense, I will do it for all the methods.

arc-mlir/src/arc-dialect.cpp
```diff
@@ -37,6 +38,91 @@ ArcDialect::ArcDialect(mlir::MLIRContext *ctx) : mlir::Dialect("arc", ctx) {
...
+  case StandardTypes::F64:
+  case StandardTypes::Integer:
+  case StandardTypes::Vector:
+  case StandardTypes::RankedTensor:
+  case StandardTypes::UnrankedTensor:
+  case StandardTypes::UnrankedMemRef:
+  case StandardTypes::MemRef:
+  case StandardTypes::Complex:
+  case StandardTypes::Tuple:
+    for (auto t : type.cast<TupleType>().getTypes()) {
```
frej: `type.cast<TupleType>().getTypes()` can't be right, none of the fallthroughs above are tuples, the cast will assert.

arc-mlir/src/include/arc/arc-dialect.h
frej: For simple things like that, which fits on one line, I'm fine with doing it inline.

arc-mlir/src/arc-dialect.cpp
frej:
```patch
diff --git a/arc-mlir/src/arc-dialect.cpp b/arc-mlir/src/arc-dialect.cpp
index cd8f967..8964665 100644
--- a/arc-mlir/src/arc-dialect.cpp
+++ b/arc-mlir/src/arc-dialect.cpp
@@ -58,15 +58,12 @@ bool isValueType(Type type) {
   case StandardTypes::UnrankedMemRef:
   case StandardTypes::MemRef:
   case StandardTypes::Complex:
+  case StandardTypes::None:
+    return true;
   case StandardTypes::Tuple:
-    for (auto t : type.cast<TupleType>().getTypes()) {
-      if (!isValueType(t)) {
+    for (auto t : type.cast<TupleType>().getTypes())
+      if (!isValueType(t))
         return false;
-      }
-    }
-    return true;
-    break;
-  case StandardTypes::None:
     return true;
   default:
     return false;
```
is probably what you want.

segeljakt: ✅

segeljakt:
@frej It should be ready now

PR #49, opened 2020-01-30
If support

frej:


arc-mlir/src/tests/ifs.mlir
```diff
@@ -0,0 +1,16 @@
...
+  func @main() {
+    %a = constant 0 : i1
+    %b = constant 3.14 : f64
+    %c = constant 0.693 : f64
+
+    "arc.if"(%a) ( {
+      "arc.block.result"(%b) : (f64) -> f64
+    },  {
+      "arc.block.result"(%c) : (f64) -> f64
+    }) : (i1) -> f64
```
segeljakt: one question, can you have nested ifs?

frej: They work fine, I added a test for that.

PR #57, opened 2020-02-12
Refactor

segeljakt:


segeljakt:
@frej This now compiles and successfully runs all tests on my Macbook 🎉

frej:
I started a build now, I'll tell you how it went tomorrow.

frej:
That one worked, now testing ~1

frej:
90b0dc3a16a49e03b2e44972207290f9fb5a523a this is needed to fix it:

```patch
diff --git a/arc-mlir/src/lib/Types.cpp b/arc-mlir/src/lib/Types.cpp
index f3b38ed..37ad89c 100644
--- a/arc-mlir/src/lib/Types.cpp
+++ b/arc-mlir/src/lib/Types.cpp
@@ -79,13 +79,13 @@ Type AppenderType::getMergeType() const { return getImpl()->mergeType; }
 Type AppenderType::parse(DialectAsmParser &parser) {
   if (parser.parseLess())
     return nullptr;
-  llvm::SMLoc loc = parser.getCurrentLocation();
+  Location loc = parser.getEncodedSourceLoc(parser.getCurrentLocation());
   mlir::Type mergeType;
   if (parser.parseType(mergeType))
     return nullptr;
   if (parser.parseGreater())
     return nullptr;
-  return AppenderType::getChecked(mergeType, loc);
+  return AppenderType::getChecked(mergeType.getContext(), mergeType, loc);
 }
```

frej:
I have tested from 89347794535ec1fc92bc9c9bbfb227cfc58e6ca9 to f5673c8af613b3be777514c792e2f3ce8778fc7b, they all build. 5e03821ddcfbabe6457d7436e77dc46556730ab7 fails to build.

segeljakt:
All builds pass now for me

segeljakt:
I created `lib/arc/` in preparation for the next dialect

segeljakt:
@frej Do you prefer keeping `.h` and `.td` files together, or in different directories? For example, keeping `.td` in `src/tblgen/`.

frej:
> @frej Do you prefer keeping `.h` and `.td` files together, or in different directories? For example, keeping `.td` in `src/tblgen/`.

Keep them together, that's the way upstream LLVM does it.

arc-mlir/src/include/arc/Ops.td
```diff
@@ -84,7 +84,7 @@ class ArcFloatUnaryOp<string mnemonic, list<OpTrait> traits = []>
     : ArcUnaryOpSameOperandAndResultType<mnemonic, traits>,
       Arguments<(ins FloatLike:$operand)>;
 
-def SinOp : ArcFloatUnaryOp<"sin"> {
```
frej: Why? Upstream uses this convention, so why should we be different?

arc-mlir/src/include/arc/Ops.td
```diff
@@ -56,7 +77,7 @@ def MakeVector
   let verifier = [{ return customVerify(); }];
 }
 
-def ArcKeep : Arc_Op<"keep", []> {
```
frej: Don't drop the Arc prefix, I think the namespace for these operations is flat, so we want our own unique prefix.

arc-mlir/src/include/arc/Ops.td
```diff
@@ -193,7 +197,7 @@ def IndexTuple : Arc_Op<"index_tuple", [NoSideEffect]> {
   let verifier = [{ return customVerify(); }];
 }
 
-def ArcBlockResult :
+def BlockResult :
```
frej: Don't drop the unique prefix

arc-mlir/src/include/arc/Ops.td
segeljakt: Ok, there were some ops before without the `Op` suffix. To be consistent I can amend so every op has an `Op` suffix. @frej WDYT?

arc-mlir/src/include/arc/Ops.td
segeljakt: Is there a rule for deciding which we want to have `Arc` as a prefix for?

arc-mlir/src/include/arc/Ops.td
frej: Do the amend to make sure we follow the convention.

arc-mlir/src/include/arc/Ops.td
frej: Looking at the other dialects, it looks like all tablegen defs has a prefix in order to not clash when a tablegen file is included.

segeljakt: True, maybe it is best to prefix everything with Arc as well?

frej: I think so

segeljakt:
I had to make a change in the `Simplify naming scheme` commit.

frej:
> I had to make a change in the `Simplify naming scheme` commit.

It still looks good to me :)

PR #58, opened 2020-02-13
Appender Operations

segeljakt:
The operations required to get appenders working. I wonder if we should / how we can generalise these to work with any kind of builder. I see two candidate solutions, one involving C++ and one involving TableGen.


frej:
The tablegen approach will probably require less future maintenance.

segeljakt:
I will try and generalize these in #65. In the meanwhile, can we merge this @frej ?

PR #63, opened 2020-02-17
WIP: Forward type inference

segeljakt:
The aim of this PR is to implement basic forward type inference, i.e. inferring the return type of an operator based on its operands.


segeljakt:
I think we can close this for now. Forward type inference might that interesting and types should generally be inferred before arc-mlir.

frej:
I agree, it's not our top priority, arc-script will do it for us.

PR #64, opened 2020-02-17
WIP: Generate documentation

segeljakt:
The aim of this PR is to generate and embed MLIR documentation into the README.


frej:
I know this is WIP, but commiting automatically generated stuff to git is not good. If you want documentation online we should cook up something using github pages and a cmake target for publishing it.

segeljakt:
That sounds like a nice plan, I will investigate

PR #65, opened 2020-02-17
Generic Builders

segeljakt:
The aim of this PR is to make builder operations generic. (e.g. `arc.merge` should work for all builder types)


segeljakt:
This should now be ready for review @frej

The PR adds a `BuilderType` as a supertype of `AppenderType`. `BuilderType` is only visible to the compiler internals, so unlike `AppenderType` there is no syntax for using it in the assembly format.

segeljakt:
One small problem, I was not able to run `clang-format`, I will try fixing it

segeljakt:
I amended with formatting

PR #67, opened 2020-02-24
Add orphan rule

segeljakt:


segeljakt:
This makes it so there will only be one commit on the `gh-pages` branch. I don't think it is necessary to version the generated files.

frej:
So that's the way to force a "force push" to the documentation area, good! That's exactly what we want.

segeljakt:
Now, the docs can be viewed at https://cda-group.github.io/arc/

PR #69, opened 2020-02-26
Add self-hosted runner and workflow for continuous integration

segeljakt:
This PR sets up a self-hosted runner for GitHub Actions. However, it seems like GitHub currently does not offer the any way to regulate whom can execute code on the machine. Soon enough, it will be possible to restrict runners to only execute code pushed by contributors within the organization:

https://github.community/t5/GitHub-Actions/Expose-self-hosted-runner-to-organization-only/m-p/44176

We could make the repository private, but I think it is best to just wait until the update.

### DIFF
--- a/.github/workflows/arc-script.yml
+++ b/.github/workflows/arc-script.yml
@@ -1,0 +1,44 @@
+name: arc-script
+
+on:
+  pull_request:
+    branches: [ mlir ]
+
+env:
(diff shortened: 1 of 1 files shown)


segeljakt:
The action will remain queued until I start the runner :smile:

segeljakt:
Until then, we can keep an eye on GitHub's changelog: https://github.blog/changelog/

PR #70, opened 2020-02-27
Frej/rust output

frej:
Changes all over the place, I think it could be good to review and merge this to avoid future conflicts.


segeljakt:
LGTM

PR #71, opened 2020-03-03
Update LLVM submodule

frej:
Changes needed:

* The MLIR cmake files now provide properties containing all the
  dialect and conversion libraries.


segeljakt:
It compiles perfectly, LGTM

PR #74, opened 2020-03-09
Add ForOp (Functional for-loop operator)

segeljakt:


segeljakt:
Currently the operator can only iterate over tensors, but we can extended it later to support range and zip iterators also. I think we can maybe wait with deriving the optimisations from the builtin loops.

arc-mlir/src/include/arc/Ops.td
```diff
@@ -26,6 +26,7 @@
 #ifndef OP_BASE
 include "mlir/IR/OpBase.td"
 #endif // OP_BASE
+/* include "mlir/Transforms/LoopLikeInterface.td" */
```
frej: This shouldn't be here

arc-mlir/src/include/arc/Ops.td
```diff
@@ -258,6 +261,62 @@ def IfOp : Arc_Op<"if", [SingleBlockImplicitTerminator<"ArcBlockResultOp">]> {
   let hasCanonicalizer = 1;
 }
 
+def ArcYieldOp : Arc_Op<"yield", [Terminator, SameOperandsAndResultType]> {
```
frej: This appears to almost the same as the `ArcBlockResultOp`, do we need both? `ArcYieldOp` is a better name so I prefer to keep the new operator. Looking at `ArcBlockResultOp` I see that it doesn't have any parent constraint, should you not require it to be a child of a loop? I think it should be possible to an 'or' constraint and use it both for if and loops.

arc-mlir/src/include/arc/Ops.td
```diff
@@ -258,6 +261,62 @@ def IfOp : Arc_Op<"if", [SingleBlockImplicitTerminator<"ArcBlockResultOp">]> {
...
+    ```
+  }];
+
+  let arguments = (ins AnyType:$result);
+  let results = (outs AnyType);
+}
+
+def ForOp : Arc_Op<"for", [
+  Linear,
+/*   DeclareOpInterfaceMethods<LoopLikeOpInterface>, */
```
frej: Forgotten comment?

arc-mlir/src/include/arc/Ops.td
segeljakt: I'll try it out

arc-mlir/src/lib/Arc/Dialect.cpp
```diff
@@ -152,6 +152,75 @@ LogicalResult IfOp::customVerify() {
   return mlir::success();
 }
 
+LogicalResult ArcYieldOp::customVerify() {
+  auto Op = this->getOperation();
+  auto ParentOp = Op->getParentOp();
+  auto Results = ParentOp->getResults();
+  auto Operands = Op->getOperands();
+
+  if (isa<IfOp>(ParentOp) || isa<ForOp>(ParentOp)) {
```
frej: `isa<IfOp>(ParentOp) ||` shouldn't this go into the next patch if we are picky?

arc-mlir/src/lib/Arc/Dialect.cpp
```diff
@@ -152,6 +152,75 @@ LogicalResult IfOp::customVerify() {
...
+      return Op->emitOpError() << "parent of yield must have same number of "
+                                  "results as the yield operands";
+    for (auto e : llvm::zip(Results, Operands)) {
+      if (std::get<0>(e).getType() != std::get<1>(e).getType())
+        return Op->emitOpError()
+               << "types mismatch between yield op and its parent";
+    }
+  } else {
+    return Op->emitOpError()
+           << "yield only terminates If, For or Parallel regions";
```
frej: "If, " to the next patch?

segeljakt:
This is ancient so I will close it

PR #76, opened 2020-03-10
Update LLVM submodule

frej:
Changes needed:

* The NonNegativeI64Attr ODS constraint has been removed and replaced
  with Confined<I64Attr, [IntNonNegative]>.

* Update tests to account for the changed wording in the error message
  due to the change above.


segeljakt:
Interesting, LGTM

PR #78, opened 2020-03-16
Update LLVM submodule

frej:
Changes needed:

  * Dialect classes are now generated with tablegen. The CMake support
    for this generates a header named ${dialect}Dialect.h.inc which
    will clash with our ArcDialect-files. So do the rename circus
    again, rename $<name>Dialect<Foo> to simply <name><foo>.

    Drop manually defined dialect classes, move dialect-specific
    declarations into extraClassDeclaration-blocks in the dialect
    .td-spec.


segeljakt:
LGTM!

PR #79, opened 2020-03-26
Update LLVM submodule

frej:
Changes needed:

  * The mlir/Analysis/Passes.h has been removed.

  * Rewrite patterns are no longer two stage. Convert the
    ConstantFoldIf rewriter to be single stage.


segeljakt:
I compiled and tested successfully, LGTM!

PR #80, opened 2020-03-26
mlir: Assume toplevel expressions return i32

frej:
This will require revisiting when the external interface is designed.


segeljakt:
Alright, LGTM!

PR #91, opened 2020-04-23
Add support for rust-native tests

frej:


frej:
> Looks great to me  
> Would it be easy or hard to extend this with support for SIMD scalars?
> https://doc.rust-lang.org/1.26.0/std/simd/index.html

It should only require us to model it in the Arc dialect and then decide how the operations and types should be converted to Rust. So I don't think it will be particularly hard nor easy.

PR #92, opened 2020-04-28
Update LLVM submodule

frej:
Changes needed:

* The symbol trait has become an OpInterface, so include the right
  header.


segeljakt:
Ok 👍

PR #98, opened 2020-05-12
Lowering of various math functions

frej:


frej:
> let bar: Bar = foo.bar(); // Approach 1
> let bar: Bar = <Foo as IntoBar>::bar(foo); // Approach 2

I think the cost for both is more or less equivalent, both require us to first look up `Foo` in the symbol table to get its type. In the first case we then have to check if there is a method called `bar` in one of the traits (this could be a scan or a hash lookup). In the second case we have to check that `Foo` implements the trait (this could be a scan or a hash lookup). What is the common way to write this in Rust? If it's the first way, I think we can assume that this resolving strategy is optimized to not require a linear scan of all implemented traits.

segeljakt:
Sorry I forgot to answer, the first way is more natural for method calls on expressions. The second way is used for static methods. https://github.com/dtolnay/quote#making-method-calls

Method calls also have a high precedence, but since it is in SSA form we probably don't need to worry about that in either case. https://doc.rust-lang.org/reference/expressions.html

PR #105, opened 2020-06-09
LLVM submodule update

frej:
Changes needed:

  * MLIR now pretty prints i1 as "constant true" / "constant false",
    so change the Scala MLIRPrinter to do the same.

  * Update tests to account for the changed printing of i1.

  * Adapt tests to changed logic for printing verification error
    notes (no note is printed when the error and note would refer to
    the same location).


segeljakt:
LGTM!

PR #106, opened 2020-06-11
Move arc-script to arc repo

segeljakt:
This PR moves arc-script into the arc repository and merges the github workflows and READMEs.


frej:
It builds and the tests pass, so LGTM!

frej:
Wait a minute! This isn't good. Running cargo build+test sets a git hook (without asking) `hooks/pre-push` which makes it impossible to push anything that's not a child of this new commit:

```
+cargo test --all
error: could not find `Cargo.toml` in `/var/tmp/arcon/arc` or any parent directory
error: failed to push some refs to 'github.com:cda-group/arc.git'
```

frej:
A better way to handle hooks is to have the real hook checked into the repo (in for example `hooks/<name>`, a documented and opt-in way to install them. Then the hook in `.git/hooks` should check that `hooks/<name>` exists before exec-ing it. That way you won't get strange crashes when you work on a branch since before the hook was created.

segeljakt:
Hooks from https://github.com/rhysd/cargo-husky should now be removed!

frej:
Much better, LGTM!

PR #109, opened 2020-06-16
Initial support for Arc and Rust structs

frej:


arc-mlir/src/include/Rust/Rust.td
```diff
@@ -288,6 +289,20 @@ def Rust_RustDependencyOp : Rust_Op<"dependency", [NoSideEffect]> {
   }];
 }
 
+def Rust_RustMakeStructOp
+    : Rust_Op<"make_struct", []> {
+  let summary = "constructs a struct";
+  let description = [{
+    "rust.structt" constructs a tuple
```
segeljakt, line 296: should this be `rust.structt` or `rust.struct`?

frej: That's a typo, will fix.

PR #110, opened 2020-06-22
Update LLVM submodule

frej:
Changes needed:

  * Upstream has refactored how block argument types get
    converted. The type converter is no longer passed to
    applyFullConversion() and for regions we have to call
    rewriter.convertRegionTypes().


segeljakt:
LGTM

PR #123, opened 2020-07-27
arc-script updates

segeljakt:
The main contribution of this PR is a symbol-table which is used to store symbolic identifiers for functions, let-expressions, and type-aliases outside of the AST. Each symbols in the symbol table additionally stores a type. This could maybe be extended with more semantic information later on.

The symbol-table is constructed during parsing (i.e., while building the AST). To this end, parsing of functions, let-expressions, and type-aliases must be separated into a declaration and definition step. For example:
```rust
LetDecl: (Ident, Box<Expr>) = <VarSymbol> <OptionalType> "=" <Box<Arg>> => {
  let (sym, ty, arg) = (<>);
  stack.push_scope();
  let id = table.insert(Decl::new(sym, ty, VarDecl));
  stack.bind_local(sym, id);
  (id, arg)
};

LetDef: ExprKind = <LetDecl> <Box<Body>> => {
  let ((id, arg), body) = (<>);
  stack.pop_scope();
  Let(id, arg, body)
};

Var: ExprKind = <VarSymbol> => {
  stack.lookup(<>)
       .map(Var)
       .unwrap_or(ExprErr)
};
```
When parsing a declaration, any symbol encountered is inserted into the symbol-table. By inserting, a unique identifier is returned as a key to the symbol. The idea is to only store the unique identifiers in the AST, and not the actual symbols, for space-efficiency. The Decl/Def distinction ensures that the function's name and parameters can be inserted into the symbol table before they are used (in the body).

The symbol is also inserted, along with its key, into a scope in the symbol-stack. The symbol-stack stores each scope and its bindings which allows resolving the symbol of any encountered variable/nominal-type/function-call to its closest binding. One additional note is that functions and type-definitions are, like in C, bound in a global namespace and variables from let-expressions in a local namespace. When binding into the local namespace, symbols are inserted into the innermost (last) scope of the stack. When binding into the global namespace, symbols are inserted into the outermost (first) scope of the stack. If a symbol with the same name already exists in that scope, a name-clash error is emitted.

A new scope is pushed onto the symbol-stack whenever a body-expression is encountered. This applies both to bodies of functions and let-expression, which means shadowing is allowed between variables.

One limitation currently is that some symbols cannot be resolved during parsing. For example, symbols of field-access expressions. This will be handled in a later PR.


segeljakt:
I realized I need to add one more commit before this potentially can be merged

segeljakt:
> WIP Arc script updates
> This commit message does not look like it is something that is finished, you should probably think about merging it into past or future patche as it is called "WIP"

Should be adressed now.

> 
> feat(dependency): add educe as a dependency
> Why is this done? The commit message doesn't say. The commit contains changes to structs and arrays, changes which the commit message is silent about and appears unrelated.

Should be adressed now with improved commit message.

> 
> style(rustfmt): run rustfmt
> As this is a not yet merged patch, at least parts of it could be merged into the previous patches.
> 

All stylistic changes have now been put into the "Cleanup" commit.

> feat(symbols): add symbol table
> We talked about how the symbol table was to be used during parsing, please document this in a comment somewhere so a new developer won't have to reverse engineer how it is used.
> 

Will do this soon.

> feat(ast): make script contain the Info struct
> I'm missing the motivation in the commit message.
> 

Squashed this into the symbol table commit.

> fix(symbols): fix identifier indices
> This looks like something that should be merged into a prevous patch
> 

Squashed this into the symbol table commit.

> style(clippy): fix clippy warnings
> as above
> 

Squashed this into the cleanup commit.

> style(utils): rename fun to f
> as above

Squashed this into the cleanup commit.

> 
> style(imports): organize imports
> as above
> 

Squashed this into the cleanup commit.

> style(grammar): create new parameters rule
> I'm missing the "why"-part in the commit message.
> 

This should've been in the symbol-table commit from the start, so I squashed it in.

> style(rustc): remove unused lifetime
> Merge this into a previous patch

Squashed this into the cleanup commit.

segeljakt:
I'm currently trying to fix one more thing. I was stupid and did a large alpha-renaming of the enum variants in d38f4f2 so I will attempt to separate this from that commit.

arc-script/src/symbols.rs
```diff
@@ -0,0 +1,148 @@
+use crate::ast::*;
+use crate::error::CompilerError;
+use DeclKind::*;
+
+/// The symbol-table is used to store symbolic identifiers for functions, let-expressions,
```
frej, line 5: Almost all non-code lines in this comment adds trailing whitespace.

segeljakt: Oops, fixed!

frej:
LGTM

PR #124, opened 2020-07-29
Attempt to fix the CI

segeljakt:
GitHub actions, please work!

### DIFF
--- a/.github/workflows/arc-mlir.yml
+++ b/.github/workflows/arc-mlir.yml
@@ -1,0 +1,40 @@
+name: arc-mlir
+
+on:
+  pull_request:
+    branches: [ mlir ]
+
+jobs:
+  build:
+    runs-on: self-hosted
+
+    steps:
+    - name: Checkout source
+      uses: actions/checkout@v2
+
+    - name: Install JDK 1.8
+      uses: actions/setup-java@v1
+      with:
+        java-version: 1.8
+
+    - name: Install Scala
+      uses: olafurpg/setup-scala@v5
+
+    - name: Install ninja-build tool
+      uses: seanmiddleditch/gha-setup-ninja@v1
+
+    - name: Update submodule
+      run: git submodule update --init --recursive
+
+    - name: Build and test arc-frontend
+      run: |
+        cd arc-frontend
+        sbt test assembly
+
+    - name: Build and test arc-mlir
+      run: |
+        cd arc-mlir
+        ./arc-mlir-build
+        ninja check-arc-mlir -C $A2M_BUILD/llvm-build/
+      env:
+        A2M_BUILD: ${{ format('{0}/build', runner.temp) }}
--- a/.github/workflows/arc-script.yml
+++ b/.github/workflows/arc-script.yml
@@ -6,7 +6,10 @@ on:
 
 env:
   CARGO_TERM_COLOR: always
-  working-directory: ./arc-script
+
+defaults:
+  run:
+    working-directory: ./arc-script
 
 jobs:
   build:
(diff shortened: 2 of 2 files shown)


.github/workflows/arc-mlir.yml
```diff
@@ -35,10 +35,9 @@ jobs:
       run: |
         cd arc-mlir
         ./arc-mlir-build
-        for COMMIT in $COMMITS; do
+        for COMMIT in $(git rev-list mlir..HEAD); do
```
frej, line 40: Add something like  `echo Testing $COMMIT` here as otherwise we have no clue which commit failed

frej, line 40: Instead of doing the loop in this script, wouldn't a cleaner version be to trigger a separate workflow doing the test of each commit? My thinking is that this action which triggers on a push only runs rev-list and then triggers a separate manually activated work flow for each commit.

Have a look at https://docs.github.com/en/actions/reference/events-that-trigger-workflows#workflow_dispatch it looks like you would have to mess around with access tokens and stuff, but it looks like it would be possible,

segeljakt: I checked the container, it seems like there is just one commit: 🤔 
```
root@48f46d5ba2ba:~/Workspace/github-actions-runner/_work/arc/arc# git log
commit 8d33340a6ab4dd4019e92a60cba30bdbe7422013 (grafted, HEAD, pull/125/merge)
Author: Frej Drejhammar <frej.drejhammar@gmail.com>
Date:   Wed Jul 29 13:32:40 2020 +0000

    Merge 4b91be891ccd97b54c994af2f62c893ce2f4e6ff into e5ffeb5b802c5535a55aa57c4d898427affaf2b6
root@48f46d5ba2ba:~/Workspace/github-actions-runner/_work/arc/arc# git status
HEAD detached at pull/125/merge
nothing to commit, working tree clean
root@48f46d5ba2ba:~/Workspace/github-actions-runner/_work/arc/arc# git branch
* (HEAD detached at pull/125/merge)
```

segeljakt: 👍 I will take a look

segeljakt: I asked on the GitHub actions board and they just sent a response: https://github.community/t/execute-job-for-all-commits-in-pull-request/125140

frej: I'm not sure how github will handle the results, but if you get a test-line for each commit it will be easier to see which commit failed.

frej: Didn't you clone with depth=0?

frej: > I asked on the GitHub actions board and they just sent a response

I only get less and less impressed by github, the more I use it. They clearly don't support a git patch-series in any meaningful way.

segeljakt: This one seems relevant https://github.community/t/run-a-workflow-for-a-specific-commit-sha/18303/10

segeljakt: and https://github.com/marketplace/actions/workflow-dispatch

segeljakt: So I guess I need to 1) Fetch the GitHub commits, 2) extract the SHA:s, 3) Post a `workflow_dispatch` events for every SHA.

segeljakt: I thought 0 = fetch everything
https://github.com/actions/checkout
```
    # Number of commits to fetch. 0 indicates all history for all branches and tags.
    # Default: 1
    fetch-depth: ''
```

segeljakt:
Is this OK to merge? @frej

.github/workflows/arc-mlir.yml
frej: You're right, I was off-by-one.

frej:
Merge away!

PR #125, opened 2020-07-29
Add support for lowering tensors

frej:
These patches add support for the tensor type in the backend


segeljakt:
Looks good to me 👍 One thing we might need to support later is to enable features for dependencies. For example, 
```
[dependencies]
ndarray = { version = "1.0.0", features = [ "blas" ] }
```
I think Cargo also supports this syntax:
```
[features]
default = [ "ndarray/blas" ]

[dependencies]
ndarray = "1.0.0"
```

frej:
> Looks good to me  One thing we might need to support later is to enable features for dependencies. For example,

That should be uncomplicated to implement when we need it.

PR #126, opened 2020-08-03
Run rustfmt to fix failing builds

segeljakt:


segeljakt:
This is very old, I will close it.

PR #128, opened 2020-08-03
Time literals, interning, and task syntax

segeljakt:
This PR brings 3 additions:
* **Time literals:** Time is often an important property in streaming systems for defining windowing computations. Often, languages express time as unsigned integers of milliseconds (or sometimes seconds), e.g., a minute is `60*1000`. In my opinion this is sometimes tough to read. For this reason, I added some syntactic sugar on top of this to maybe improve the readability. You can now write `60s`, `60000ms`, `60000000us`, `60000000000ns`, `1h/60`, `1min`.
* **Interning:** All strings generated by the compiler are now stored in a singular data structure, and identical strings are stored at the same memory address. These strings (`SymbolName`) can then be accessed using integer keys (`SymbolKey`). This means that no strings are stored in the AST, which should improve performance and reduce the overall memory consumption.
* **Task syntax:** This commit adds syntax for defining tasks. A task is a low-level primitive which resembles a node in the dataflow graph. It will be possible to implement any higher order function as a task. Tasks will have a number of functions that can optionally be implemented to configure how the task operates. I will try define these in the next PR.


segeljakt:
This PR is large so I don't expect you to review everything. There should be no warnings or test failures but I'm not sure all commits will make sense. I can spend some more time cleaning it up if you want.

frej (approved):
I trust you!

segeljakt:
I forgot to pull mlir before rebasing. Now there should be no merge conflicts.

frej:
> I forgot to pull mlir before rebasing. Now there should be no merge conflicts.

Good, then we know that the MLIR updates haven't broken anything, or at least not something we test.

PR #135, opened 2020-08-12
Make tests go green

frej:
This patch has been in #128 for over a week and as I think this is what's needed to make the tests go green, try to merge this separately.


frej:
Bad Frej! You should merge to `mlir` not `master`!

PR #140, opened 2020-10-06
LLVM submodule update

frej:
Changes needed:

  * atan is now in the standard dialect, use it instead of our own
    implementation in the arc dialect.

  * The tablegen-generated dialect headers now include the dialect
    namespace, so stop double wrapping the inclusion.

  * The library implementing the standard operations has been renamed
    from MLIRStandardOps to MLIRStandard.

  * Operation::index() noew returns a uint64_t instead of an APInt.

  * A call to mlir::enableGlobalDialectRegistry(true) is now needed to
    automatically register all dialects with the LLVMContext.


segeljakt (approved):
Looks good to me! I will eventually add `atan` and other trig-ops to arc-script.

PR #143, opened 2020-10-14
rust: Fix new warnings produced by Clang 11

frej:
Switching from Clang 10 to 11 triggered new warnings about overridden
methods not marked with `override`.


segeljakt (approved):
LGTM

PR #146, opened 2020-11-02
LLVM submodule update

frej:
Changes needed:

  * Change Rust call operator to use OpBuilderDAG instead of the
    deprecated OpBuilder.


segeljakt (approved):
LGTM

PR #153, opened 2020-11-05
Compile cmake with cargo

segeljakt:
This is a WIP to merge the compilation and testing of the arc-script and arc-mlir compilers. We should probably discuss more to find the best solution.


segeljakt:
I will close this because the problem got solved in #156

PR #155, opened 2020-11-10
WIP - Arc-Script rewrite

segeljakt:
This PR is WIP and currently introduces the following changes:

* **Adds a module system which takes inspiration from Rust.**

It will be possible to execute the compiler with `arc-script run foo.arc` which compiles `foo.arc` and any dependencies it may have. Dependencies follow a basic version of Rust's system:
```
foo.arc      # ::
bar/
  mod.rs     # ::bar
  baz/
    mod.rs   # ::bar::baz
  qux/
    mod.rs   # ::bar::qux
```
You can also omit the file `arc-script run`, in this case it will search upwards in the file system for a directory which contains an `Arcon.toml` and `src/main.rs`, and compile the `src/main.rs`:
```
Arcon.toml
src/
  foo.arc      # ::
  bar/
    mod.rs     # ::bar
    baz/
      mod.rs   # ::bar::baz
    qux/
      mod.rs   # ::bar::qux
```
* **Adds four new IRs**

Previously the parser directly produced an AST. The parser now instead produces a parse tree for each source file. The parse trees are then combined into a single HIR (High-Level IR) which is a set of tables for looking up definitions through symbolic paths. This should give a lot more flexibility for adding new features. After having constructed the HIR, it is then evaluated into a static DFG (Dataflow Graph). The DFG and HIR are then both lowered into MLIR (Multi-Level IR). Each IR can be pretty-printed (i.e., code-generation for MLIR). In addition to the IRs, an Info struct is kept on the side which stores options, information about names, paths, files, diagnostics, and types. Optionally, the HIR and DFG can also be lowered directly into a Rust-IR (using the `syn`-crate) for Rust-code generation.


frej:
The CI build error is because dd10cb365c3b51183e1060c4ede77f0cfde1ccbc updates (backdates?) the LLVM-submodule to a much older version.

frej:
The frej/arc-script branch is this PR but rebased onto the current mlir.

segeljakt:
If this succeeds, then feel free to merge 👀

frej (approved):
LGTM and also builds for me.

PR #158, opened 2020-11-16
arc-script: Fix syntax errors in the generated MLIR

frej:
* Use the custom syntax for arc constants

* Use the custom syntax for arc binary operators

* Arc syntax for a boolean true is `1 : i1`, false is `0 : i1`

* Apart from booleans, Arc uses the signed/unsigned ints


segeljakt (approved):
LGTM

PR #162, opened 2020-11-18
arc-script: Output correct MLIR function arguments

frej:
Function arguments require both an indentifier and a type.


segeljakt (approved):
LGTM

PR #163, opened 2020-11-20
mlir: Change signature for arc.block.result

frej:
Make arc.block.result have the same signature as std.return, ie
"arc.block.result"(%val) : (<type>) -> ()


segeljakt (approved):
Looks good to me!

PR #164, opened 2020-11-24
LLVM submodule update

frej:
Changes needed:

  * Function declarations inside an Arc module cannot be public, see
    [1].

[1] https://llvm.discourse.group/t/rfc-symbol-definition-declaration-x-visibility-checks/2140


segeljakt (approved):
👍

PR #171, opened 2021-01-19
LLVM submodule update

frej:
Changes needed:

  * I64EnumAttrs are no longer printed in double quotes, so change the
    arc-to-mlir frontend and the tests to assume the same.


segeljakt:
LGTM, I'll update arc-script too

segeljakt (approved):

PR #172, opened 2021-01-19
Support for Arcon's state Value

frej:


segeljakt (approved):
It looks great to me 👍

PR #173, opened 2021-01-21
arc-mlir: Implement operations for Arcon's state Appender

frej:


segeljakt (approved):
Nice work, looks good to me 👍

PR #179, opened 2021-02-02
Change Path representation + minor improvements

segeljakt:
This PR changes how paths are represented in Arc.

Previously paths were stored as interned vectors:
```rust
struct PathId(usize);
struct PathBuf {
    segments: Vec<Name>
}
struct PathInterner {
    id_to_path: Vec<PathBuf>,
    path_to_id: HashMap<PathBuf, PathId>,
}
```
Multiple identical paths could reuse the same memory, but multiple paths with the same parent path couldn't. It was also tough to append new path segments, since you needed to resolve the `PathBuf`, clone it, append the new segment, and intern the new path. Now paths are stored as interned linked lists:
```rust
struct PathId(usize);
struct PathBuf {
    pred: Option<PathId>,
    name: Name,
}
struct PathInterner {
    id_to_path: Vec<PathBuf>,
    path_to_id: HashMap<PathBuf, PathId>,
}
```
This makes it a lot easier to append new segments. Probably the one downside is that prepending segments is costly, but this is rarely needed.

To squash bugs while implementing this I also added more methods to debug the internal data structures of the compiler.


frej (approved):
The code LGTM, but the long explanation that is in the PR should really go in the commit as that's where we will look, if we are investigating some future strangeness using `git blame` and `git log`.

segeljakt:
Sorry for committing more after your commented. I undid the commit. You do not need to update your review.

segeljakt:
I will try and improve the documentation is subsequent commits.

PR #181, opened 2021-02-03
Add bf16 and f16

segeljakt:
This PR adds two more floating point types using the [`half`](https://crates.io/crates/half) Rust-crate.


arc-script/arc-script-core/src/compiler/ast/from/parse/error.rs
```diff
@@ -1,7 +1,7 @@
 use crate::compiler::ast;
 use crate::compiler::ast::from::parse::lexer::Token;
 use crate::compiler::info::diags::{Diagnostic, Error};
-use crate::compiler::info::files::{FileId, Loc, ByteIndex};
+use crate::compiler::info::files::{ByteIndex, FileId, Loc};
```
frej, line 4: Does this belong to this commit?

frej (approved):

arc-script/arc-script-core/src/compiler/ast/from/parse/error.rs
segeljakt: Not sure how that made it in, but I undid it now :)

segeljakt:
> If my calculations are correct, when this baby uses `f88`... you're gonna see some serious shit.

arc-script/arc-script-core/src/compiler/ast/from/parse/error.rs
frej: A second pair of eyes is allways good.

PR #183, opened 2021-02-09
LLVM submodule update

frej:
Changes needed:

  * StringAttr::get() has switched the order of the arguments.

  * Methods/functions returning a LogicalResult now expects the result
    to be checked -- make sure we do that.


segeljakt:
LGTM

PR #190, opened 2021-02-15
LLVM submodule update

frej:
Changes needed:

 * The math operations in the standard dialect have moved to their own
   `math` dialect. Update tests, Arc-to-Rust lowering and dialect
   loading in arc-mlir.

 * MLIRContext::getDialectRegistry() now returns a const value. Change
   registration logic in arc-mlir to avoid warning.


segeljakt (approved):
LGTM, interesting to see a new dialec.

PR #192, opened 2021-02-16
Arc script updates and bugfixes

segeljakt:
This PR fixes some issues which in result makes more tests pass. This can possibly be useful for MLIR.


arc-script/arc-script-core/src/compiler/ast/display.rs
```diff
@@ -396,7 +396,6 @@ impl<'i> Display for Pretty<'i, ast::BinOp, Context<'_>> {
             ast::BinOpKind::Leq  => write!(f, " <= "),
             ast::BinOpKind::Or   => write!(f, " or "),
             ast::BinOpKind::And  => write!(f, " and "),
-            ast::BinOpKind::Xor  => write!(f, " xor "),
```
frej, line 399: Actually we could support xor for booleans quite simply if you want it. The code generator would just transform  `(A xor B)` to `((A == !B) || (!A == B))`

frej (approved):
LGTM, the only thing is if you really want do drop xor for bools.

arc-script/arc-script-core/src/compiler/ast/display.rs
frej: Or, as booleans are `i1` in MLIR, you could just use the bxor, the result is the same.

segeljakt: Interesting, then I might add it back.

segeljakt: Is it `arc.xor` or `std.bxor`?

frej: `std.xor`: https://mlir.llvm.org/docs/Dialects/Standard/#stdxor-xorop

arc-script/arc-script-test/insta/src/expect-pass/binops.arc
```diff
@@ -265,6 +265,8 @@ fun main() {
 #CHECK: {{%[^ ]+}} = and {{%[^ ]+}}, {{%[^ ]+}} : i1
     let lor: bool = c_bool or c_bool in
 #CHECK: {{%[^ ]+}} = or {{%[^ ]+}}, {{%[^ ]+}} : i1
+    let lxor: bool = c_bool xor c_bool in
+#CHECK: {{%[^ ]+}} = arc.xor {{%[^ ]+}}, {{%[^ ]+}} : i1
```
frej, line 269: It should be std.xor, arc.xor is not defined for `i1`, only `uiX` and `siX`.

frej (changes requested):

frej (approved):

PR #195, opened 2021-02-19
arc script mlir fixes

frej:
* Lower math.powf properly
* Output NaNs and +/- inf correctly
* Run insta/src/expect-pass/binops.arc with arc-to-rust

### DIFF
--- a/arc-mlir/src/include/Rust/RustPrinterStream.h
+++ b/arc-mlir/src/include/Rust/RustPrinterStream.h
@@ -75,6 +75,7 @@ public:
 
   void flush() {
     OS << "#![allow(non_snake_case)]\n#![allow(non_camel_case_types)]\n";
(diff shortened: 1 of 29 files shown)


segeljakt:
LGTM, looks like you found a bug in arc-script with `pow`. I can address it in a PR.

frej:
> LGTM, looks like you found a bug in arc-script with `pow`. I can address it in a PR.

Are you sure, I thought it was that you have to use `powf`/`powi` depending on the type?

segeljakt:
>Are you sure, I thought it was that you have to use powf/powi depending on the type? I thought it was just a bug in the MLIR backend, why didn't it fail before?

I have not yet added so the tests check the generated code but I probably should :eyes:

segeljakt (approved):
LGTM

PR #198, opened 2021-02-23
The next level

segeljakt:
This adds another module `query` which implements an SQL-like language using the `sqlparser` crate.

The language is preliminarily named "Arq" and will translate into Arc-ASTs (through a Stream-Relational Algebra).


frej (approved):
LGTM

PR #201, opened 2021-03-02
Frej/backend switch

frej:


segeljakt (approved):
Nice, looks good to me!

segeljakt (approved):
Looks very good, we can also maybe later add so that insta runs with the new backend

segeljakt:
👍

PR #202, opened 2021-03-02
Arc-Script Updates

segeljakt:


segeljakt:
this was working on my machine before I uploaded it 👀

segeljakt:
ok, now I see what can go wrong with `compiletest_rs`. If you got duplicate build artifacts (`cargo check` and `cargo test`), then `compiletest_rs` cannot distinguish between them and might pick the wrong one (`cargo check`) by chance

the solution is to do `cargo clean` before `cargo test`

frej (approved):
LGTM if you figure out why it's crashing :)

segeljakt:
I am 99% sure this will work, let's see

arc-mlir/src/CMakeLists.txt
```diff
@@ -25,4 +25,6 @@ get_filename_component(ARC_SCRIPT_SRC_DIR "../../arc-script"
   REALPATH BASE_DIR ${CMAKE_CURRENT_SOURCE_DIR} CACHE)
 
 add_custom_target(arc-check DEPENDS check-arc-mlir
-  COMMAND ${ARC_CARGO_BIN}/arc-cargo test --manifest-path=${ARC_SCRIPT_SRC_DIR}/Cargo.toml)
+  COMMAND cd ${ARC_SCRIPT_SRC_DIR}; cargo clean
```
frej, line 28: This should be `arc-cargo clean` or else you won't clean the target dir you run `arc-cargo test` in. Otherwise lo

frej (approved):
Apart from the comment, this LGTM.

PR #206, opened 2021-03-15
LLVM submodule update

frej:
Changes needed:

 * Forward declarations are no longer sufficent for referencing the
   custom TypeStorage structs, from the dialect implementations. Move
   the contents of {Arc,Rust}/Types.cpp into the corresponding
   Dialect.cpp.

 * Canonicalization no longer preserves the ordering of constants (or
   we have just been lucky before), so switch
   arc-mlir/src/tests/opts/canonicalization.mlir to use CHECK-DAG.


segeljakt (approved):
LGTM!

PR #207, opened 2021-03-15
arc-script: Provide nicer errors when build script fails

frej:
Instead of just asserting with an "Internal Compiler Error", include
the name of the file which crashed the compiler as we can't be sure
that the crashing compiler reports it.


segeljakt (approved):
LGTM

PR #208, opened 2021-03-18
Frej/arc script mlir support

frej:


segeljakt:
Very nice. I wonder if it was a mistake of me to pull in `f16` and `bf16`. I have not found a way to do operations on them as of yet 🤔  do you have any ideas?

segeljakt (approved):

frej:
> I wonder if it was a mistake of me to pull in f16 and bf16. I have not found a way to do operations on them as of yet thinking do you have any ideas?

I assumed you had some crate supporting operations on them in mind when you introduced them :)

MLIR, and some of the GPU backends have support for them, so perhaps we should just wait for a crate to appear.

PR #215, opened 2021-03-31
arc-script: Report backend crashes using compile_error!()

frej:
As we don't want the build script to terminate as soon as a backend
crashes but instead continue to generate other test cases, we report
internal errors by creating a dummy output file containing only a
compile_error!().


segeljakt (approved):
LGTM 🐰 🥚

PR #219, opened 2021-04-13
arc-script: Include file and line number when reporting ICEs

frej:
Usa a panic hook to capture the panic location and payload when an
internal compiler error is triggered.


segeljakt (approved):
Nice, LGTM

PR #220, opened 2021-04-13
Port mlir-related tests to use the same output format as the arcon backend

frej:


segeljakt:
The error might be related to rustc, I updated rustc to the latest version now, let's see if it works.

segeljakt (approved):
LGTM

PR #221, opened 2021-04-14
Fix website and add a language reference

segeljakt:
This should fix the problems we have been having with github pages


frej (commented):
LGTM, I will have a look at the language ref. when it's rendered.

PR #223, opened 2021-04-22
Enum operations

frej:
Constant folding of `enum_access` and `enum_check` is still left to do, but these operations can be reviewed and merged piecemeal.


segeljakt (approved):
Awesome! LGTM

PR #228, opened 2021-05-07
Make it possible to access enums by path

segeljakt:
Now you can use the enum-api as:
```rust
    #[arcorn::rewrite]
    enum Foo {
        Bar(i32)
    }

    let foo = enwrap!(Foo::Bar, 5);
    assert!(is!(Foo::Bar, foo));
    assert_eq!(unwrap!(Foo::Bar, foo), 5);
```

Note: The old API is still supported. Eventually I will migrate to this in arc-script.


frej (approved):
LGTM!

PR #229, opened 2021-05-07
Export paste crate from arcorn to fix bug

segeljakt:


segeljakt:
Previously the macro would bug because it could not find the `paste` macro.

frej (approved):
Now it works, so LGTM!

PR #234, opened 2021-05-26
Task conversion

frej:
Implement support for producing arcorn tasks from MLIR


segeljakt (approved):
very nice, maybe we can try generate this from arc-script

PR #235, opened 2021-05-31
LLVM submodule update

frej:
Changes needed:

  * MLIR constants are no longer materialized in the order of use, so
    a test case had to be rewritten to use CHECK-DAG to match the
    constants and then use CHECK to check for the correct values at
    the time of use.


segeljakt (approved):
LGTM

PR #236, opened 2021-05-31
Rework compiler

segeljakt:
This PR reworks the core and the api of the arc-script compiler.


segeljakt:
@frej it seems like the `arcorn::state` thing is going nuts. In Arctime, state-types are not hardcoded so we do not need to include this module anymore. Arcorn might provide them as intrinsics to Arcon however but we don't need to think about this right now. For the moment, Arctime is the default runtime for the purpose of experimentation.

frej:
LGTM, go ahead and merge this.

PR #237, opened 2021-06-03
Allow master to be latest released nightly Rust

frej:
The only thing that stops master from being compiled with the latest
released nightly version of Rust is the use of the experimental
`or_patterns` feature.

This patch enables `or_patterns` and disables the deprecated
`unstable_features` lint for the relevant crate.


segeljakt (approved):
LGTM

PR #238, opened 2021-06-03
Check that a build does not modify version controlled files

frej:
This is done to keep us from committing changes to Cargo.toml-files
without committing the corresponding changes to Cargo.lock.


segeljakt (approved):
LGTM, this will save some headache 👍

PR #239, opened 2021-06-03
Add MLIR dialect experiments directory

segeljakt:
The plan for this directory is to test out MLIR dialects and to try link their generated code into Rust.


mlir-dialect-experiments/exec-mlir-from-rust/Makefile
```diff
@@ -0,0 +1,10 @@
+all:
+	clang -c mlir-src/hello.c \
+		    -o mlir-target/hello.o && \
+	llvm-ar rc \
```
frej, line 4: This isn't good makefile style:

* Make knows how to build libraries, you just have to override `AR` if you want `llvm-ar` to be used.
* Make also knows how to build a `.o` from a `.c`, just override `CC` to `clang`
* This should be split into mutliple rules, one for building the `.o` one for creating the library.

mlir-dialect-experiments/exec-mlir-from-rust/build.rs
```diff
@@ -0,0 +1,6 @@
+use std::process::Command;
+
```
frej, line 2: Apart from the make style issues, the Rust build script should not invoke make. The CMake part should be responsible for building the library, when it exists corrosionm should invoke cargo which then can assume that the library exists.

frej (changes requested):

PR #242, opened 2021-06-17
Fix so enum-ops can be prefixed with module name

segeljakt:
Now it is possible to have three kinds of paths in ops:

```rust
unwrap!(my_module::MyEnum::MyVariant, x);
unwrap!(MyEnum::MyVariant, x);
unwrap!(MyVariant, x);
```

One day I will try to generalise the solution to any kind of path. It's
tricky with Rust-macros.


frej (approved):
It's mostly greek to me, but I can believe it :)

segeljakt:
Declarative macros are like regex, they are write-only.

PR #244, opened 2021-06-21
LLVM submodule update

frej:
Changes needed:

  * Remove previously removed rust.crate_end operation from tests.


segeljakt (approved):
LGTM

PR #248, opened 2021-06-21
Fix continuous integration for `cargo insta`

segeljakt:
This commit makes it so changes in snapshots, as a result of running
`cargo insta test`, will be catched and reported during CI.


frej (approved):
Good!

PR #251, opened 2021-06-23
Ifs without values

frej:


segeljakt (approved):
LGTM, interesting but scary 🙀

PR #252, opened 2021-06-24
Fix MLIR codegen of unit type

segeljakt:
Before there was a bug which caused this to generate:

```
foo(...) : (...) ->
```

Now we get:

```
foo(...) : (...) -> ()
```


frej (approved):
LGTM.

PR #253, opened 2021-06-24
arc-script: Don't omit `arc.block.results` in MLIR backend

frej:
The blocks in an `arc.if` have to have terminators even if the
`arc.if` does not return a value, so generate an `arc.block.result"()
: () -> ()` in that case.


segeljakt (approved):
LGTM

PR #254, opened 2021-06-28
Annihilate the `unit` type from MLIR codegen

segeljakt:
MLIR does not like unit types and therefore this commit rewrites all `unit`
types and values into the `void` type (i.e., `()`). If a function takes
parameters which are `unit`, those parameters are also removed.

### DIFF
--- a/arc-script/arc-script-core/src/compiler/ast/repr.rs
+++ b/arc-script/arc-script-core/src/compiler/ast/repr.rs
@@ -482,6 +482,7 @@ pub enum ScalarKind {
     U32,
     U64,
     Str,
+    // Eliminated when lowering to HIR
     Unit,
     DateTime,
     Duration,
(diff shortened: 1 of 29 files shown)


frej (commented):
I think this may be a bit too brutal as for enums we will still need `none` as the type for variants without values, `()` isn't  valid type for a variant.

Remove the `XFAIL` from arc-script-test/compile/src/tests/expect_pass/{enum*,option}.arc}` to see the problem.

frej:
The failing test is due to `arc.enum_access "crate_Foo_Bar" in (%x_1 : !arc.enum<crate_Foo_Bar : none>) : ()` it has to be `arc.enum_access "crate_Foo_Bar" in (%x_1 : !arc.enum<crate_Foo_Bar : none>) : none`. 

But instead of fixing that, I think you should change the test as it doesn't make sense to first check `arc.enum_check (%x_1 : !arc.enum<crate_Foo_Bar : none>) is "crate_Foo_Bar" : i1` and then do an access of the same variant.

frej (approved):
LGTM!

PR #258, opened 2021-06-30
Test artifacts generated by Arc-MLIR in compile tests

segeljakt:


segeljakt:
The arc-script compiler error can be fixed with `export PATH="PATH:$path/to/arc-mlir/build/llvm-build/bin/"`, I should maybe add this to the CI 🤔

frej:
> I should maybe add this to the CI thinking

That's probably a good move :)

segeljakt:
The best course of action here might be to merge the CI of `arc-script` with `arc-mlir`. This should ensure that the `arc-mlir` binary always exists before running `arc-script`.

arc-script/arc-script-test/compile/src/tests/mod.rs
```diff
@@ -25,7 +26,31 @@ mod expect_pass {
     test!(t16, "src/tests/expect_pass/structs.rs");
     test!(t17, "src/tests/expect_pass/basic_by.rs");
     test!(t18, "src/tests/expect_pass/sort_fields.rs");
-//     test!(t19, "src/tests/expect_pass/unit_fun.rs");
+    test!(t19, "src/tests/expect_pass/unit_fun.rs");
+}
+
+/// Check that artifacts generated by Arc-MLIR compile successfully with Rust.
+mod expect_pass_opt {
```
frej, line 33: Having hard-coded paths isn't good. Fixing a test that used to not work with the MLIR backend should only require it to be moved from the `expect_mlir_fail_todo` directory to `expect_pass`. I for sure will never remember to update this list too.

If we're moving towards merging the MLIR and arc-script CI pipelines, why can't we use Lit to run these tests? That way we can have one `RUN`-line for each backend. Alternatively can't we get `build.rs` to scan for markers (just like Lit) in the files?

segeljakt: > Having hard-coded paths aren't good. Fixing a test that used to fail with the MLIR backend should only require it to be moved from the expect_mlir_fail_todo directory to expect_pass. I for sure will never remember to update this list too.

I could make it glob instead if you would prefer it.

> If we're moving towards merging the MLIR and arc-script CI pipelines, why can't we use Lit to run these tests? That way we can have one RUN-line for each backend.  Alternatively can't we get build.rs to scan for markers (just like Lit) in the files?

I am up for both, the first seems like the best approach to me. Can Lit be used for integration tests? For example if we want to include the generated sources in another file.

frej: I would prefer a glob.

> I am up for both, the first seems like the best approach to me. Can Lit be used for integration tests? For example if we want to include the generated sources in another file.

Lit is used for all tests in the LLVM repo (at least I think so), so yes. When I test the generated Rust in the MLIR-pipeline I have a shellscript which produces a small wrapper crate around the generated code and then runs `cargo test` on it. The shell script is what is called from the Lit `RUN`-lines.

The shellscript could easily be extended to also  include another file (given on the command line) in the crate which could bring in custom tests cases (Rust assertions) or select a different wrapper crate.

segeljakt: Would it be ok to address this in the next PR?

frej: Merge away!

PR #259, opened 2021-06-30
Add channels to arctime

segeljakt:
Channels are asynchronous ADTs which support two operations:

* push/1 - push element into channel
* pull/0 - pull element from channel

These are not used by arcorn currently, but could eventually become
useful.

### DIFF
--- a/arctime/src/channel.rs
+++ b/arctime/src/channel.rs
@@ -1,0 +1,123 @@
+#![allow(clippy::type_complexity)]
+#![allow(deprecated)]
+#![allow(dead_code)]
+
+use arrayvec::ArrayVec;
+use kompact::component::AbstractComponent;
+use kompact::prelude::*;
+use time::*;
+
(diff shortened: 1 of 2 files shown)


arctime/src/channel.rs
```diff
@@ -0,0 +1,119 @@
...
+    fn push(&self, data: T) {
+        self.0.tell(ChannelMessage::Push(data));
+    }
+
+    fn pull(&self) -> KFuture<T> {
+        self.0
+            .ask_with(|promise| ChannelMessage::Pull(Ask::new(promise, ())))
+    }
+}
+
```
frej, line 98: Shouldn't this and the following function have some kind of `#![test]`-marker?

frej (approved):

arctime/src/channel.rs
segeljakt: Good point 👍

PR #262, opened 2021-07-01
Remove f16 and bf16

segeljakt:
Neither `f16` or `bf16` are supported by Rust. Previously we were
parsing and printing them using the `half` crate, but were unable to use
them for any kind of arithmetic. As the types are currently serve no
use, this commit removes them. We might add them back if Rust ever
adds support for more floating point types.


frej (approved):
LGTM

PR #270, opened 2021-07-16
mlir-rust: Correct Rust output for unit valued enum variants

frej:
When the `unit` type is used to declare an enum variant it should be
written as `unit`, but when used as the type of a variable or the
return type of a function it should be `()`.


segeljakt (approved):
LGTM

PR #275, opened 2021-07-19
Implement panic for the mlir arc dialect

frej:


segeljakt (approved):
LGTM. Once we have strings we could add so the message can be a variable.

frej:
> LGTM. Once we have strings we could add so the message can be a variable.

Or a format string, and the variables would be arguments. That way we won't have to instantiate a string constant for static messages.

PR #279, opened 2021-07-20
Compile tasks into FSMs

segeljakt:


segeljakt:
Closing this, I will implement the transformation in OCaml before Rust

Issue #277, opened 2021-07-19
Cache artifacts generated by CI


segeljakt:
Caching of artifacts which are generated during CI could speedup compilation and testing times.

We could try the following (suggested by Frej):

1. Increase `ccache` to ~20G
2. Cache the target-directory used by the `arc-cargo`-script.

I will put this issue here until I have resolved it

frej:
For 1 I would start with checking that the CI doesn't zap the ccache cache directory by default.

PR #280, opened 2021-07-20
Increase ccache maxsize to 20G

segeljakt:
Addresses #277

### DIFF
--- a/.github/persist.sh
+++ b/.github/persist.sh
@@ -1,0 +1,11 @@
+#!/bin/bash
+if [[ -d "${PERSIST_DIR}/ccache-cachedir" ]]; then
+    echo "The Ccache directory exists at ${PERSIST_DIR}/ccache-cachedir"
+else
+    echo "Creating Ccache directory at ${PERSIST_DIR}/ccache-cachedir"
+    mkdir -p ${PERSIST_DIR}/ccache-cachedir
+    envsubst > ${PERSIST_DIR}/ccache-config <<EOF
+    max_size = 20G
+    cache_dir = ${PERSIST_DIR}/ccache-cachedir
+EOF
+fi
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -4,12 +4,14 @@ on:
   pull_request:
     branches: [ master ]
 
-env:
-  CARGO_TERM_COLOR: always
-
 jobs:
   build:
     runs-on: self-hosted
+
+    env:
+      CARGO_TERM_COLOR: always
+      CCACHE_CONFIGPATH: /root/actions-runner/persist/ccache-config
+      PERSIST_DIR: /root/actions-runner/persist/
 
     steps:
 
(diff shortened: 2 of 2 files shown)


arc-mlir/arc-mlir-build
```diff
@@ -41,6 +41,7 @@ A2M_ASSERTS="${A2M_ASSERTS:+-DLLVM_ENABLE_ASSERTIONS=ON}"
 
 # Do we want ccache?
 A2M_CCACHE="${A2M_CCACHE:+-DLLVM_CCACHE_BUILD=1}"
+A2M_CCACHE_MAXSIZE="${A2M_CCACHE:+-DLLVM_CCACHE_MAXSIZE=20G}"
```
frej, line 44: I'd prefer if this was only done on the CI-machine, users have their own preferences for the location and size of their ccache setup. If I did this I'd:

1. Figure out a path on the CI-machine which is writeable by the runner, but not zapped by it between runs and call this `<PERSIST_DIR>`

2. Create a shell script:
  ```
 #!/bin/bash
mkdir -p ${PERSIST_DIR}/ccache-cachedir
envsubst > ${PERSIST_DIR}/ccache-config <<EOF
max_size = 20G
cache_dir = ${PERSIST_DIR}/ccache-cachedir
EOF
  ```
3. In the configuration for the workflow environment I would set:
 ```
PERSIST_DIR="<PERSIST_DIR>"
CCACHE_CONFIGPATH="${PERSIST_DIR}/ccache-config"
```
And then configure the workflow to run the shell script before it runs `arc-build`

**Review by frej — CHANGES_REQUESTED** [review]
Ccache configuration should not be dictated by arc-mlir-build, see the inline comment.

**Comment by frej** [comment]
To speed up the build set `BUILD_FLAVOUR=Release` `A2M_ASSERTS="1"` when executing `arc-mlir-build`. That will drastically speed up linking, but still give check assertions. The less understandable backtraces won't be a problem as we are not debugging on the CI machine.

File: .github/workflows/test.yml [hunk 2]
```diff
@@ -21,8 +21,14 @@ jobs:
     - name: Update submodule
       run: git submodule update --init --recursive
 
+    - name: Persist artifacts
+      run: ./.github/persist.sh
+      env:
+        PERSIST_DIR: /root/actions-runner/persist/
+        CCACHE_CONFIGPATH=/root/actions-runner/persist/ccache-config
```
frej, line 28: Check if there is a global environment for all actions, this way `CCACHE_CONFIGPATH` will have to be repeated for the "Build and test arc-mlir and arc-script" action.

The syntax looks fishy: Why is it `:` in one assignment and `=` in the other.

segeljakt: This explains why the syntax highlighting didnt work 😅

.github/workflows/test.yml
```diff
@@ -21,8 +21,14 @@ jobs:
...
+    - name: Persist artifacts
+      run: ./.github/persist.sh
+      env:
+        PERSIST_DIR: /root/actions-runner/persist/
+        CCACHE_CONFIGPATH: /root/actions-runner/persist/ccache-config
+
       # Arc-MLIR specific tests
-    - name: Build and test arc-mlir
+    - name: Build and test arc-mlir and arc-script
       run: |
```
frej, line 32: I suspect `CCACHE_CONFIGPATH` won't be set for this action.

frej (approved):
If this works I'm happy with the implementation.

.github/persist.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+mkdir -p ${PERSIST_DIR}/ccache-cachedir
```
frej, line 2: A late addition, wrap the existing body in:
```
if [[ -d "${PERSIST_DIR}/ccache-cachedir" ]] then
    echo "The Ccache directory exists at ${PERSIST_DIR}/ccache-cachedir"
else
   echo "Creating Ccache directory at ${PERSIST_DIR}/ccache-cachedir"
  ... old stuff ...
fi
```
That way we can check the logs to verify that the directory isn't removed by the runner

PR #283, opened 2021-07-21
mlir-rust: Verify the type signature of event handlers

frej:
This is not a full check, but it is enough to avoid segfaults in later
stages and errors undetected here will trigger type errors when the
generated Rust is compiled. Considering that types will be checked at
the arc-script level, this should be sufficient to detect bugs in the
arc-script MLIR-generator.


segeljakt (approved):
LGTM

PR #285, opened 2021-07-21
mlir: Verify that enum_access isn't used with none-typed variants

frej:
Trying to retrieve the value of a none-typed variant does not make
sense, so let the verifier refuse such a constructs.


segeljakt (approved):
LGTM

PR #288, opened 2021-07-22
Catch and report panic messages

segeljakt:
Catches and reports panic messages in addition to payload and location
whenever a panic occurs in the compiler.


frej (approved):
LGTM

PR #291, opened 2021-07-23
Do CI in stages

frej:


segeljakt (approved):
Cool, LGTM

.github/workflows/test.yml
```diff
@@ -23,10 +23,68 @@ jobs:
...
+      env:
+        A2M_BUILD: ${{ format('{0}/build', runner.temp) }}
+
+    - name: "Run arc-script test: check --examples"
+      run: ./.github/continuous-integration.sh cargo check --examples
+      env:
+        A2M_BUILD: ${{ format('{0}/build', runner.temp) }}
+
+    - name: "Run arc-script test: check --benches"
+      run: ./.github/continuous-integration.sh cargo fuzz run parse -- -runs=10000 -only-ascii
```
segeljakt, line 80: I think there is a bug here between `cargo check --benches` and `cargo fuzz`.

segeljakt (changes requested):

.github/workflows/test.yml
frej: Two things in the old script became one in this one, fixed.

segeljakt (approved):

segeljakt:
LGTM

PR #295, opened 2021-07-26
build-system: End CI runs with a report on cache usage

frej:
Dump Sccache and Ccache statistics at the end of the CI run.

### DIFF
--- a/.github/continuous-integration.sh
+++ b/.github/continuous-integration.sh
@@ -37,6 +37,8 @@ fi
 function check-ccache {
     echo "Ccache statistics:"
     ccache -s
+    echo "Cccache statistics:"
+    sccache --show-stats
 }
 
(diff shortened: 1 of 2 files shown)


segeljakt (approved):
LGTM, cool

segeljakt:
One thing I'm wondering, the environment variables inside the script are only set in the CI. For example `PERSIST_DIR`. The script is very useful for reproducing errors locally. Could we make so the variables have a default value, maybe relative to `$A2M_BUILD` so we can run the script outside of CI without having to set them ourselves?

frej:
> What do you think about making it so the variables have a default value, maybe relative to `$A2M_BUILD` so we can run the script outside of CI without having to set them ourselves?

That could be useful, I can do it in another PR.

PR #296, opened 2021-07-26
Add unsafe impl for generated components

segeljakt:


arc-script/arc-script-api/arcorn/arctime/macros/src/tasks.rs
```diff
@@ -224,6 +224,10 @@ pub(crate) fn rewrite(attr: syn::AttributeArgs, item: syn::ItemMod) -> pm::Token
                 #(pub #state_name: State<#state_type>,)*
             }
 
+            // Allows the component to be sent across threads even if it contains fields which are 
+            // not safely sendable. The code generator is required to generate thread-safe code.
+            unsafe impl Send for #task_name {}
+
```
frej, line 230: Is it possible to, in a debug build, make the runtime complain loudly if the code generator fails to maintain the invariants?

frej (approved):
LGTM

segeljakt:
We can type check the event type at compile time in Rust to ensure it is Send, while the state type does not need to be

PR #297, opened 2021-07-28
mlir: Don't forget to copy string terminator

frej:
We cannot assume that `hex` is initialized to zero, so copy the string
terminator too when moving the exponent.


segeljakt (approved):
LGTM

Wow 😄  low level stuff

PR #299, opened 2021-07-28
Convert sharable types into sendable types (and vice versa)

segeljakt:
For once, I present to you some really dank code


frej (approved):
LGTM. Is it limitations in the macro system that prevents you from inspecting the types of the struct fields? At the MLIR-level I don't think we'll have any problems generating the annotations by just looking at the types.

frej:
Oh, the tests failed, I guess we'll have to implement the annotations in the MLIR code generator.

segeljakt:
> LGTM. Is it limitations in the macro system that prevents you from inspecting the types of the struct fields? At the MLIR-level I don't think we'll have any problems generating the annotations by just looking at the types.

It is possible to inspect the types syntactically and map them to their categories. Primitives should probably be hard coded into the code generator. The distinction between algebraic and dynamic types is only about performance. We for example do not need box non-recursive types `Box<String>`. Still, maybe it is a good idea to box them to simplify the generated code.

Currently we have that `impl From<Operable> for Sendable` and `impl From<Sendable> for Operable`. Where

```rust
trait From<T> {
    fn from(x: T) -> Self;
}
```

This means multiple conversions could exist per type. I am going to change this to a new trait `Convert`:

```rust
trait Convert {
    type T;
    fn convert(self) -> Self::T;
}
```

And have `impl Convert for Sendable { type T = Operable }` and `impl Convert for Operable { type T = Sendable }`. This ensures only a single conversion exists per type. Because of this, we can also retrieve the corresponding type after conversion for any type. This however requires that all types, including primitives, implement the trait. I think it can be a good approach however since it gives us full control over which types are allowed and not allowed to run on our system.

segeljakt:
Made some dank mods to the conversion code, now it should be way easier to generate

segeljakt:
We can bring arc-mlir up-to-date next week

segeljakt:
Now we can have some fun and send recursive types over the network OwO'

segeljakt:
I could undo the last commit if you do not like it (added it to make the generated code less verbose).

arc-mlir/src/include/Rust/RustPrinterStream.h
```diff
@@ -139,8 +139,8 @@ class RustPrinterStream {
       if (t.isa<FunctionType>() || t.isa<types::RustEnumType>() ||
           t.isa<types::RustStructType>() || t.isa<types::RustTensorType>() ||
           t.isa<types::RustTupleType>())
-        return "v" + std::to_string(id) + ".clone()";
-      return "v" + std::to_string(id);
+        return "arcorn::val!(v" + std::to_string(id) + ")";
+      return "arcorn::val!(v" + std::to_string(id) + ")";
```
frej: With `acorn::val!` used for all types, special-casing the non-primitive types is no longer needed, so drop the `if`.

frej (approved):
LGTM when you fix the inline comment.

> To make the generated code less verbose I removed the arcorn:: prefix from paths

That's an improvement, so go for it.

arc-mlir/src/include/Rust/RustPrinterStream.h
segeljakt: Ok, it's an abstraction with "hopefully no regret" 😅

PR #303, opened 2021-08-02
Add support for opaque types that are passed on to Rust

frej:


segeljakt (approved):
LGTM, this is some seriously dope code!

PR #304, opened 2021-08-02
Add support for strings

segeljakt:
This commit adds support for strings. Strings are represented as abstract data types. The implementation reuses Rust's `std::string::String`. What do you think, is this a good representation? From what I see there are four possible representations:

* UTF8 (`std::string::string`)
* ASCII (`std::vec::Vec<u8>`)
* Lists (`std::collections::LinkedList<char>`)
* Ropes (https://crates.io/crates/ropey)

Lists are horrible without lazy evaluation, but ropes could be useful since they are both efficient and completely functional. In my implementation I wrapped UTF8 inside a reference counter. If two strings are used in an operation like concatenation, then a new string is allocated for the result instead of mutating one of the strings.


frej:
> What do you think, is this a good strategy?

ASCII-only does not cut it in the 21:st century :) A linked list representation is horribly memory ineffiicent and Ropes has too much overhead for the small strings we can expect in a program (format strings, status messages etc). So I'd say that `std::string::string` is the right compromise.

segeljakt:
I agree we should go for `std::string::String`. Should it be ephemeral (mutable cell) or persistent (clone on write)?

frej:
How will it look at the arc-script level? As arc-script is SSA, CoW will probably create the least number of surprises, but it is, as you say, not very efficient when doing incremental changes. We could use CoW for now, when/if it is an actual performance problem, move it to a native arc-dialect type and implement an optimization which converts operations into destructive updates where it can be done safely.

segeljakt:
At the arc-script level it would look like:
```
val x = "hello worl";
val y = x.append('d');
```
Eventually when the type system is more advanced we could create syntactic sugar
```
val x = "hello worl";
val y = x + 'd';
```

arc-script/arc-script-core/src/compiler/mlir/display.rs
```diff
@@ -414,7 +414,7 @@ pretty! {
             mlir::ScalarKind::DateTime => write!(w, "ui64"),
             mlir::ScalarKind::Duration => write!(w, "ui64"),
             mlir::ScalarKind::Str      => crate::todo!(),
-            mlir::ScalarKind::Char     => crate::todo!(),
+            mlir::ScalarKind::Char     => write!(w, "char"),
```
frej: Should this be in this commit? You change it in a later commit.

frej (approved):
LGMT

arc-script/arc-script-core/src/compiler/mlir/display.rs
segeljakt: Oops, the first commit should also probably come in later

PR #305, opened 2021-08-03
mlir-rust: Allow the name used for functions to be overridden

frej:
The name used by a function in the generated Rust can be overridden by
setting the attribute `arc.rust_name` on the function.


segeljakt (approved):
Looks good to me!

PR #307, opened 2021-08-04
Restructure arc-script project

segeljakt:


frej (changes requested):
The llvm-submodule shouldn't be reverted to an older version (or updated at all). You have probably forgotten to run `git submodule update` and then just added everything.

Otherwise it looks good to me.

segeljakt:
I was too eager with `git add -u`

frej (approved):
LGTM

PR #312, opened 2021-08-08
Rewrite Arc-Script compiler in OCaml

segeljakt:
This is a PR which adds another compiler for arc-script, written in OCaml (reusing code I wrote in a previous course I took, given by David Broman). In contrast to the Rust-compiler, this compiler is meant for quick experimental prototyping, mainly for research publications. It is not intended to be used by end-users. The following features are not a priority:
* ~~Syntax - The compiler does not have a fancy parser/printer. Instead it just derives an S-Expression parser/printer for the AST.~~ (Update: The parser is now at the same level as the Rust implementation and uses the [Menhir](http://gallium.inria.fr/~fpottier/menhir/) parser generator)
* Performance - Performance tricks like replacing strings with integers is not a priority. The compiler will be entirely functional.
* Modules - A program is just a single file.
* Error handling - If it is possible I will maintain line information in the AST to track where errors occur but anything beyond that, such as error recovery, is probably not going to happen.
* User interface - LSP/REPL/CLI will not be supported. The compiler will just take a file, read, compile it, and produce output.


segeljakt:
@frej just an FYI. I changed this PR so it goes into `frej/WIP/receive-op`.

PR #318, opened 2021-08-23
mlir: Fix crash in verification of `arc.block.result`

frej:
If the number of results of the parent `arc.if` did not match the
number of operands to `arc.block.result` the verifier would crash with
a segfault during type verification. The fix is to check that the
number of results match before trying to verify the types.


segeljakt (approved):
Looks good to me!

PR #319, opened 2021-08-23
Allow returns from inside `if`s

frej:


segeljakt (approved):
Interesting.. Looks great to me!

PR #321, opened 2021-10-05
LLVM submodule update

frej:
Changes needed to the arc-script implementation:

* Upstream has changed all symbol table look ups to use StringAttrs
  instead of StringRefs, so we have to make the corresponding change.

* AsmParser::parseString() no longer expects a StringRef, so switch to
  a std::string.

* The operation printer now prints out the name of the operation
  before calling the user-defined printer, so remove the name from the
  custom printers.


segeljakt (approved):
LGTM, sorry for late review

PR #326, opened 2021-11-04
LLVM submodule update

frej:
Upstream has retired the traditional methods for doing lowering using
ConversionPattern. The new recommended way to implement lowerings is
now to use OpConversionPattern. For arc-mlir this means that we have
to update all lowerings to derive from OpConversionPattern and use
OpConversionPattern::Adaptor to access the operator operands.


segeljakt (approved):
Looks great to me :)

PR #336, opened 2021-11-24
Add MLIR tests

segeljakt:
Add tests for
* Vector ADT
* Closures
* Enums which contain variants of the same name (but with different types)
* Nested ifs
* Infinite loops
* Nested infinite loops
* Printing


frej (approved):
LGTM, but shouldn't you call all functions you define and compile (using the `-rustinclude` mechanism)? As long as the backend generates something that's syntactically valid, the `closure`, `enum`, `if` and `loop` tests will pass. I should probably have added that the first time around for the existing tests.

segeljakt:
I can fix it both for the new and old tests

frej:
> I can fix it both for the new and old tests

Perfect!

PR #342, opened 2021-12-09
mlir: Support lowering of std.select to Rust

frej:
std.select-operations are sometimes introduced by canonicalization, so
we need to support them. As Rust does not have a ternary if, we simply
lower the select to an if.


segeljakt (approved):
LGTM

PR #343, opened 2021-12-09
arc-mlir: Move generated pass definitions to header

frej:
With the planned addition of additional passes for the Arc dialect we
don't want to have to remember the boilerplate inclusion, so move it
to a shared header.


segeljakt (approved):
LGTM

PR #344, opened 2021-12-10
rust-mlir: Fix bug in handling of aliased types in type definitions

frej:
Fix a bug where the type of a field/variant of a struct/enum would not
use the aliased type.


segeljakt (approved):
LGTM

PR #345, opened 2021-12-13
LLVM submodule update

frej:
Changes needed:

  * Arguments and results of operators used to be accessible using a
    method named `<name>()`, the accessor is now named `get<Name>()`.


segeljakt (approved):
LGTM

PR #347, opened 2021-12-30
Rewrite compiler for Arc-Lang (previously Arc-Script) into OCaml

segeljakt:
This PR adds a compiler for Arc-Lang, written in OCaml. It also updates the README and documentation.


frej:
How much work would it be to break up this into logical chunks that would still allow us to bisect it? Is it possible at all? If not, the right thing is to just squash all of them, as we're not going to be helped (during future debugging) by an incremental change that only says "Update <foo>".

arc-mlir/arc-mlir-build
```diff
@@ -60,6 +72,8 @@ function llvm-cmake {
       return
     fi
 
+    dune-build
```
frej, line 75: As far as I can tell, this won't rebuild the OCaml parts when source files have changed. It should go in `llvm-build` or even better, figure out if there is a way to build Dune-based things from inside CMake, like we do with the cargo parts.

segeljakt: I am quite lost when it comes to CMake 👀 I can move it into `llvm-build` but I'm not sure that I can figure out the CMake stuff

segeljakt:
I can show what I mean 👀

segeljakt:
I think we need to squash the changes of `arc-{lang, runtime, codegen, script}` and CI into one commit sadly... The changes to `arc-runtime` affects both `arc-lang` and `arc-script` :(

frej:
> In my opinion, maybe it is best to squash because the changes either introduce something completely new or overhaul old functionality.

That's what I suspected, go for the squash.

frej (approved):
LGTM

PR #348, opened 2022-01-10
LLVM submodule update

frej:
Changes needed:

  * Operator regions used to be accessible using a method named
  `<name>()`, the accessor is now named `get<Name>()`.


segeljakt (approved):
LGTM

PR #349, opened 2022-01-11
Passes converting between SCF and simple branches

frej:
Implement transform passes which translate between structured control flow and branches between basic blocks.


segeljakt:
Very cool :O LGTM!

segeljakt (approved):

PR #350, opened 2022-01-14
Avoid crash with implicit module

frej:


frej:
Dropped test case, the diagnostics verification seems to go haywire when errors are emitted for line 0.

segeljakt (approved):
Perfect, LGTM

PR #351, opened 2022-01-17
LLVM submodule update

frej:
Changes needed:

* The Identifier class has been removed from the C++ API and
   StringAttr is used instead.


segeljakt (approved):
LGTM

PR #353, opened 2022-01-20
arc-mlir: Fix name of arc.receive stream parameter

frej:
The stream parameter to the receive operation should be named a
'source' and not 'sink'.


segeljakt (approved):
LGTM

PR #358, opened 2022-02-07
LLVM submodule update

frej:
Changes needed:

 * The Tablegen OpTrait has been replaced with Trait.

 * The standard control flow instructions such as for example BranchOp
   has been moved to a mlir::cf namespace.

 * std.select has moved to arith.select.

 * Changes to ConstantOp handling makes it necessary to only try to
   constant fold arith::ConstantOp values for arc.make_vector.


segeljakt (approved):
LGTM

PR #361, opened 2022-02-14
LLVM submodule update

frej:
Changes needed:

 * Old-style custom parser hooks have been removed upstream. Port the
   trickier syntax to the newer alternatives and switch simple cases to
   use ODS syntax definitions.


segeljakt (approved):
LGTM

PR #363, opened 2022-02-15
mlir: Fix references to functions with arc.rust_name attributes

frej:
The name given by the arc.rust_name attribute was not used when a
constant referring to the function was constructed.


segeljakt (approved):
LGTM

PR #364, opened 2022-02-15
rust-mlir: Construct function references using function!()

frej:
Using the new runtime library, function references must be created
using `function!()`.


segeljakt (approved):
LGTM

PR #366, opened 2022-02-18
Add Arc-Report to the repository + add steps for it in GitHub actions

segeljakt:
It's good to have the report in the repository so it is synced to code changes.


segeljakt:
I could not install LaTeX on the runner (which was running ubuntu 16.04) so I had to create a new runner. I made it so that this runner (unlike the previous one) does not execute as root. This required modifying paths in the GitHub actions workflow.

frej (approved):
LGTM

PR #367, opened 2022-02-18
Update CI setup instructions

segeljakt:
I updated the instructions to include latex, opam, OCaml and dune. Also, I configured so that the runner is not run as root. Together with this PR I created a new runner using these instructions.


frej (approved):
LGTM

PR #372, opened 2022-02-22
Fix the URL for GitHub Pages

segeljakt:
Previously it was

https://cda-group.github.io/arc/html

Now it should be

https://cda-group.github.io/arc/


frej (approved):
LGTM

PR #380, opened 2022-02-24
Allow trailing commas in all macros

segeljakt:


frej (approved):
LGTM, I guess we'll test it when we get the tests in frej/bad-tests working and merge them?

PR #381, opened 2022-02-25
Arc-report updates

segeljakt:


segeljakt:
After this I will bring the Overleaf doc up-to-date. Seif wants to contribute 👀

.github/continuous-integration.sh
```diff
@@ -40,7 +40,7 @@ if [[ -d "${SCCACHE_DIR}" ]]; then
     echo "It contains $(du -hs ${SCCACHE_DIR} | cut -f1)"
 else
     echo "Creating Sccache directory at ${SCCACHE_DIR}"
-    mkdir -p ${SCCACHE_DIR}
+    sudo mkdir -p ${SCCACHE_DIR}
```
frej, line 43: This hung before and is fishy. If the build doesn't run as root and you create the sccache dir as root, the build will not be allowed to write to it (unless you change permissions or ownership). Also there should be no need for sudo as `${SCCACHE_DIR}` is in the arc-runners home directory (are files/directories in the home directory owned by root due to a previous run where the user was still root?).

frej (changes requested):
Suspicious about the use of sudo in the script.

I haven't reviewed the text parts, that's easier to do on the built document, but that shouldn't delay merging it.

.github/continuous-integration.sh
segeljakt: Hmm, before it did not compile without sudo. I can try remove it and see what happens

segeljakt: wait, huh, I can't find these changes

frej: Me neither, looks like I reviewed something stale.

frej (approved):

.github/continuous-integration.sh
frej: It's gone now, so LGTM.

PR #383, opened 2022-02-28
LLVM submodule update

frej:
Changes needed:

  * Upstream has changed the error message when a tensor is declared
    as something else or a tensor of the wrong type. Fixed by simply
    updating the expected errors.


segeljakt (approved):
LGTM

PR #384, opened 2022-03-01
Update documentation

segeljakt:
I added documentation for Arc-MLIR. It can probably be improved a lot. @frej how do you generate arc-mlir docs? Perhaps we could remove the grammar here if the generated one is not too bad.


frej:
>how do you generate arc-mlir docs?

No clue. When we first introduced MLIR you did something in the documentation/website deploy script to generate the documentation from the `.td`-files.

segeljakt:
Here https://github.com/cda-group/arc/pull/64

PR #385, opened 2022-03-02
arc-lang: Hack around ICE in dune

segeljakt:
Concerns this issue: https://github.com/ocaml/dune/issues/5498

Instead of specifying `--build-dir <dir>`, the script now does `cd <dir>`


segeljakt:
This problem is solved by down-grading to dune version 2.8.4.

PR #386, opened 2022-03-07
LLVM submodule update

frej:
Changes needed:

 * Upstream has renamed the Standard dialect to the Func dialect,
   adapt to the new name in arc-mlir and arc-rust tools and libraries.


segeljakt (approved):
LGTM

PR #387, opened 2022-03-08
arc: Add compiler driver

segeljakt:


arc
```diff
@@ -0,0 +1,113 @@
...
+            break
+            ;;
+        *)
+            INPUT_FILE=$1
+            ;;
+    esac
+    shift
+done
+
+CRATE_NAME=$(basename $INPUT_FILE)-crate
```
frej, line 42: Quote the RHS or else things will break when `$INPUT_FILE` contains a path with a space in it.

`CRATE_NAME="$(basename "$INPUT_FILE")-crate"`

This applies to all variables in the script.

Should the crate name contain the extension of the input file?

arc
```diff
@@ -0,0 +1,113 @@
...
+    echo "arc-lang not found. Please run 'make' in the root directory."
+    exit 1
+fi
+
+if [ ! -x $(which arc-mlir) ]; then
+    echo "arc-mlir not found. Please run 'make' in the root directory."
+    exit 1
+fi
+
+if [ ! -f $ARC_RUNTIME_DIR ]; then
```
frej, line 62: I can't see that `ARC_RUNTIME_DIR` is set by the script, is this intentional?

If our build system installs `arc`, that should be enough to find the path to the other binaries. It should be enough to give a full path to `arc` for the other binaries to be found, i.e. `/foo/bar/bin/arc` should be enough.

arc
```diff
@@ -0,0 +1,163 @@
...
+if [ -z "$INPUT_FILE" ]; then
+    usage
+fi
+
+if [ -z "$CRATE_DIR" ]; then
+    CRATE_DIR="$(dirname $INPUT_FILE)/$CRATE_NAME"
+fi
+
+ARC_DIR=$(realpath $(dirname "$(which "$0")")/)
+
```
frej, line 88:
```shell
mkdir /tmp/a /tmp/b
cat >> /tmp/a/foo <<EOF
#!/bin/bash
echo "Realpath says: $(realpath $(dirname "$(which "$0")"))"
echo "Readlink says: $(dirname "$(readlink -f "$(which "$0")")")"
EOF
chmod 755 /tmp/a/foo
ln -s /tmp/a/foo /tmp/b/foo
```
Then try to run `/tmp/b/foo`:
```shell
$ /tmp/b/foo 
Realpath says: /tmp/b
Readlink says: /tmp/a
```
and you'll see that realpath gets it wrong. Check [how it can be done in a way that is known to work both on Linux, MacOs and even Wintendo](https://github.com/frej/fast-export/blob/dbb8158527719e67874e0bba37970e48e1aaedfb/hg-fast-export.sh#L6).

arc
```diff
@@ -0,0 +1,163 @@
...
+    if [ ! -f "$ARC_LANG" ]; then
+        echo "Could not find arc-lang. Did you run './build'?"
+        exit 1
+    fi
+fi
+
+if [ -z "$ARC_MLIR" ]; then
+    ARC_MLIR="$ARC_DIR/arc-mlir/build/llvm-build/bin/arc-mlir"
+    if [ ! -f "$ARC_LANG" ]; then
+        echo "Could not find arc-mlir. Did you run './build'?"
```
frej, line 100:
```
Did you run \'$ARC_DIR/build\'?"
```

arc
```diff
@@ -0,0 +1,163 @@
...
+        exit 1
+    fi
+fi
+
+if [ -z "$ARC_RUNTIME_DIR" ]; then
+    ARC_RUNTIME_DIR="$ARC_DIR/arc-runtime/"
+fi
+
+INPUT_FILE="$(realpath $INPUT_FILE)"
+CRATE_DIR="$(realpath $CRATE_DIR)"
```
frej, line 110: Missing quoting.
Why are you using `realpath` here? You have already checked that the input file exists on a (perhaps convoluted) path? `CRATE_DIR` is also derived from `INPUT_FILE` so if you insist on `realpath`, why not do it once before defining `CRATE_DIR`?

arc
```diff
@@ -0,0 +1,163 @@
...
+arc-runtime = { version = "=0.0.0", path = "$ARC_RUNTIME_DIR", features = ["legacy"] }
+hexf        = { version = "0.2.1" }
+ndarray     = { version = "0.13.0" }
+prost       = { version = "0.7.0" }
+
+[profile.dev]
+debug = false
+EOF
+
+debug "Cargo.toml: $(cat $CRATE_DIR/Cargo.toml)"
```
frej, line 146: Missing quoting

arc
```diff
@@ -0,0 +1,163 @@
...
+
+echo '#![feature(unboxed_closures)]' > "$CRATE_MAIN_FILE"
+$ARC_LANG "$INPUT_FILE" | $ARC_MLIR "$@" -arc-to-rust -inline-rust -arc-lang-runtime >> "$CRATE_MAIN_FILE"
```
frej, line 149: ditto

arc
```diff
@@ -0,0 +1,163 @@
...
+        ;;
+    build)
+        cargo build --manifest-path "$CRATE_TOML_FILE"
+        ;;
+    run)
+        cargo run --manifest-path "$CRATE_TOML_FILE"
+        ;;
+esac
+
+debug "Generated code: $(cat $CRATE_MAIN_FILE)"
```
frej, line 163: This could probably overflow the maximum size of a command line argument.

You could do:
```
 [ "$DEBUG" ] && echo "Generated code" && cat $"CRATE_MAIN_FILE"
```

frej (changes requested):
The structure is sound, but there are some problems in the line comments.

segeljakt:
Good comments. I think all should be addressed now.

arc
```diff
@@ -0,0 +1,178 @@
+#!/bin/bash
+
+# set -e
```
frej, line 3: You're not checking the result of all commands, so errors should be catched to avoid having the driver continue when something went wrong.

frej (commented):
Looks good, just enable `set -e`, that way we won't try to run cargo when the code generator failed.

arc
segeljakt: Woops, fixed now!

frej (approved):
LGTM

PR #389, opened 2022-03-11
rust-mlir: Fix bug in rust annotation handling

frej:
The "rust.annotation" function attribute did not produce the
annotation to be output. Correct the pretty printer to check for the
correct attribute.


segeljakt (approved):
LGTM!

PR #391, opened 2022-03-14
Add more functionality to compiler driver

segeljakt:


frej (approved):
LGTM

frej:
BTW, why doesn't this trigger all three CI tests?

frej:
> Do we want it to trigger them? The tests currently don't have a dependency on the `arc` file

They will after #392.

PR #392, opened 2022-03-14
Share Cargo.toml template between compiler driver and test driver

frej:
Move the `Cargo.toml` template to its own file and let CMake insert it
into arc-mlir-rust-test and arc. Also move the compiler driver from
the top-level directoiry into the bin directory like all other tools.

### DIFF
--- a/arc-lang/etc/Cargo.toml.template
+++ b/arc-lang/etc/Cargo.toml.template
@@ -1,0 +1,17 @@
+[package]
+name        = "${CRATE_NAME}"
+version     = "0.0.0"
+authors     = ["The arc-mlir testing gnome"]
+edition     = "2021"
+license     = "MIT"
+description = "Generic MLIR-to-Rust test case"
+
+[dependencies]
+arc-runtime         = { version = "=0.0.0", path = "${ARC_RUNTIME_SOURCE_DIR}", features = ["legacy"] }
+hexf                = { version = "0.2.1" }
+ndarray             = { version = "0.13.0" }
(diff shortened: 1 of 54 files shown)


arc-lang/etc/Cargo.toml.template
```diff
@@ -0,0 +1,17 @@
...
+authors     = ["The arc-mlir testing gnome"]
+edition     = "2021"
+license     = "MIT"
+description = "Generic MLIR-to-Rust test case"
+
+[dependencies]
+arc-runtime         = { version = "=0.0.0", path = "${ARC_RUNTIME_SOURCE_DIR}", features = ["legacy"] }
+hexf                = { version = "0.2.1" }
+ndarray             = { version = "0.13.0" }
+prost               = { version = "0.7.0" }
```
segeljakt, line 13: I think we should remove these dependencies. I can do it in another PR.

frej: Don't we need arc-runtime?

segeljakt (approved):
LGTM, nice work

arc-lang/etc/Cargo.toml.template
segeljakt: Ah, I thought Github would mark the lines. We should only need arc-runtime. It can re-export the other dependencies. I tried it but got a problem with `serde`, so maybe we will still need serde in the dependency list. I can try to fix it.

frej: Ok, I'll merge this then.

PR #393, opened 2022-03-14
Add more Arc-Lang functionality

segeljakt:
More commits are coming


.gitignore
```diff
@@ -151,6 +151,8 @@ history.txt
 
 arc-script/**/template/
 
+*.crate
+
```
frej, line 155: If you happen to create a crate somewhere in the source tree in a directory where Lit looks for tests, you will be spammed with a lot of inconclusive tests. In that case it's easier to remove them by just asking Git for what fouled the source tree.

arc-lang/examples/strings.arc
```diff
@@ -0,0 +1,31 @@
+# RUN: arc-lang %s | arc-mlir-rust-test %t - -rustinclude %s.rust-tests
+# RUN: arc-lang %s | arc-mlir-rust-test %t-canon - -rustinclude %s.rust-tests -canonicalize
```
frej, line 2: Use
```
# RUN: arc -o %t run %s -- -rustinclude %s.rust-tests
# RUN: arc -o %t-canon run %s -- -rustinclude %s.rust-tests -canonicalize
```

And you'll avoid all the crashes from `arc-lang` not finding its runtime libraries.

arc-lang/src/pretty_mlir.ml
```diff
@@ -80,10 +80,15 @@ and pr_type t ctx =
...
       pr_angle (pr_list (pr_field pr_type) vts) ctx;
-  | Mlir.TAdt (x, ts) ->
+  | Mlir.TAdt x ->
       pr "!arc.adt";
       pr "<";
-      pr_path x ctx;
+      pr_quote (pr_path x) ctx;
+      pr ">";
+  | Mlir.TAdtPar (x, ts) ->
+      pr "!arc.adt_template";
```
frej, line 89: The MLIR type is `!arc.generic_adt`.

arc-mlir/src/tests/arc-to-rust/loops.mlir.rust-tests
```diff
@@ -1,14 +1,15 @@
...
   use hexf::*;
   #[rewrite(main)]
   #[test]
   fn it_works() {
-    assert_eq!(call!(a_while_loop(0, 5, 3)), 13);
-    assert_eq!(call!(a_while_loop_with_a_break_in_before(0, 5, 1)), 2);
-    assert_eq!(call!(a_while_loop_with_a_break_in_after(0, 10, 3)), 4);
-    assert_eq!(call!(a_while_loop_with_a_return_in_before(0, 5, 1)), 2);
-    assert_eq!(call!(a_while_loop_with_a_return_in_after(0, 10, 3)), 4);
+    // assert_eq!(call!(a_while_loop(0, 5, 3)), 13);
```
frej, line 9: This looks fishy, if none of the tests work, why isn't the tests XFAIL?

frej (changes requested):
Some fishyness in the tests, and the wrong output for `arc.generic_adt` but otherwise this looks good

arc-mlir/src/tests/arc-to-rust/loops.mlir.rust-tests
segeljakt: With `call!(...)` it's not possible to call functions defined in Arc-Lang from Rust, only the other way around 🤔. The stuff inside `loops.mlir` is still compiled and checked.

frej: Ah, makes sense, but in that case why doesn't, for example, the `arc-cmpf` and `arc-cmpi` tests fail in the same way?

segeljakt: Or wait, I tested it now and it compiles. It's possible to `call!(...)` but only on functions defined in arc-lang

segeljakt: I reverted the change 👍

frej: I should apparently trust my nose :)

PR #394, opened 2022-03-21
LLVM submodule update

frej:
Changes needed:

 * There is no longer a way to define a custom verifer in tablegen
   using `verifier`. Instead a verifier has to use the default
   signature, and its existance is indicated by `hasVerifier`.

 * Upstream has split out AttrDef/TypeDef and pattern constructs from
   OpBase.td, so we have to include `mlir/IR/PatternBase.td` when
   defining rewrite patterns using Tablegen.

 * Adapt to upstream refactoring and moving function anc call
   interfaces to their own headers.

 * The attribute used for the type of a function has been renamed to
   `function_type` from `type` for operators using the function
   interface. So we have to do the corresponding rename for the Rust
   dialect and update the lowering of Arc to Rust.

### DIFF
--- a/arc-mlir/src/include/Arc/Arc.td
+++ b/arc-mlir/src/include/Arc/Arc.td
@@ -153,7 +153,7 @@ def MakeVectorOp
   let results = (outs 1DTensorOf<[AnyType]>);
   let hasCanonicalizer = 1;
   let extraClassDeclaration = [{ LogicalResult customVerify(); }];
-  let verifier = [{ return customVerify(); }];
+  let hasVerifier = 1;
 }
 
 def ArcKeepOp : Arc_Op<"keep", []> {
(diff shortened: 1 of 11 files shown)


segeljakt (approved):
Well done, LGTM. One question, if the changes are this big, is it worth always keeping up with the upstream, or can we at some point pin the upstream version?

frej:
Pinning to an upstream version is a good way to slowly slide into irrelevance. There are already too many interesting projects which say "unfortunately this requires an old LLVM version, but we are working on a forward port" and are never heard from again. By not staying up to date we would not benefit from upstream improvements. If we find bugs, upstream would not be interested in fixing them unless we could show that they are present in HEAD.

By syncing with upstream once a week we stay relevant and by doing it weekly there is a limited number of upstream commits to go through to find what has broken our code.

PR #395, opened 2022-03-21
rust-mlir: Lowering of arith.cmpi "eq"/"ne" for i1

frej:


segeljakt (approved):
It looks good to me. Though one thing, in Arc-Lang I convert `A != B` into `not (A == B)`. Since we don't have `not` in MLIR, I could probably generate `not X` as `X == false` or `X != true`.

frej:
> Since we don't have not in MLIR, I could probably generate not X as X == false or X != true.

We have `not`, it's just not spelled "not", if you want a truth value and not flow-control you can use xor:

```MLIR
func @not_i1(%arg0: i1) -> i1 {
  %0 = arith.constant 1 : i1
  %1 = arith.xori %arg0, %0 : i1
  return %1 : i1
}
``` 

or you can use select:

```MLIR
func @not_select_i1(%arg0: i1) -> i1 {
  %0 = arith.constant 0 : i1
  %1 = arith.constant 1 : i1
  %2 = arith.select %arg0, %0, %1 : i1
  return %2 : i1
}

```
check #396.

PR #399, opened 2022-03-24
tests: Speed up repeated test runs

frej:
By not removing the work directory we can speed up repeated tests
runs. The main benefit is due to the preservation of the Cargo target
directory, as Cargo is smart enough to avoid recompiling if the Rust
output from arc-mlir has not changed.


segeljakt (approved):
LGTM

Issue #398, opened 2022-03-23
Support nested extern types


segeljakt:
I do not think it is currently possible to have an extern type that has extern type parameters. For example, this MLIR:
```
!arc.adt<"Vec<!arc.adt<"Vec<i32>">">
```
Will in the worst case not parse, since the inner quotes are not escaped. In the best case, it will generate this Rust which does not compile:
```
Vec<!arc.adt<"Vec<i32>">
```

I think a solution is to separate the type name from the type parameters:
```
!arc.adt<"Vec", [!arc.adt<"Vec", [i32]>]>
```
This will make it easier to generate this Rust:
```
Vec<Vec<i32>>
```

@frej Is it possible?

PR #402, opened 2022-03-29
MLIR Support for Nested ADTs

frej:
The support for nested ADT types in MLIR requires a runtime without `features = ["legacy"]`, but as that fails many other tests, this patch series and #398 will have to wait until the legacyectomy is done.


segeljakt:
LGTM

PR #406, opened 2022-04-19
LLVM submodule update

frej:
Changes needed:

 * Upstream has moved mlir::FuncOp to mlir::func::FuncOp, change
   our code accordingly.

 * Upstream moved enum attribute definitions from OpBase.td to
   EnumAttr.td, include EnumAttr.td where we use them.


segeljakt (approved):
LGTM

PR #408, opened 2022-04-21
Make GitHub actions build releases

segeljakt:


segeljakt:
Hard to say if this will work. There is a risk that the GitHub actions runner runs out of space. If that's the case, then we need to use our custom runner to build everything. To build for macOS we need Docker.

.github/continuous-integration.sh
```diff
@@ -63,6 +63,40 @@ function run-runtime-tests {
...
+    opam install -y dune core menhir
+}
+
+function install-ubuntu-packages {
+    sudo apt update
+    sudo apt install software-properties-common
+    add-apt-repository ppa:avsm/ppa
+    sudo apt update
+    sudo apt install -y clang libssl-dev ninja-build wget make curl opam
+    $(install-ocaml)
```
frej, line 91: Why `$(foo)` and not just `foo`? I'm pretty sure you don't want to evaluate whatever the cmake build system writes to stdout.

.github/workflows/test.yml
```diff
@@ -62,3 +63,89 @@ jobs:
...
+      run: |
+        package="arc-${{needs.prepare-release.outputs.ARC_VERSION}}-${{matrix.target}}"
+
+        mkdir -p "$package"/{bin,share}
+
+        mv README.md "$package"
+        mv arc-mlir/build/llvm-build/bin/{arc,arc-lang,arc-mlir} "$package/bin"
+        mv arc-runtime "$package/share"
+
+        tar cvf "$package.zip" "$package"
```
frej, line 139: This isn't nice, naming an uncompressed tar ball `.zip`.

.github/workflows/test.yml
```diff
@@ -62,3 +63,89 @@ jobs:
...
+    - name: Create package
+      shell: bash
+      run: |
+        package="arc-${{needs.prepare-release.outputs.ARC_VERSION}}-${{matrix.target}}"
+
+        mkdir -p "$package"/{bin,share}
+
+        mv README.md "$package"
+        mv arc-mlir/build/llvm-build/bin/{arc,arc-lang,arc-mlir} "$package/bin"
+        mv arc-runtime "$package/share"
```
frej, line 137: Does the `arc`-script handle this out of the box?

.github/continuous-integration.sh
segeljakt: Hmm, I think I should just have parentheses

segeljakt: So it creates a new subshell

.github/workflows/test.yml
segeljakt: Good point, I think we need it to find the `arc-mlir`, `arc-lang` binaries, and the `arc-runtime` source code. How do we do it? Should we check the paths `/usr/local/share/arc-runtime/` and `/usr/local/bin/{arc-mlir,arc-lang}`?

segeljakt: Oh, it looks like you fixed it already

https://github.com/cda-group/arc/blob/9b845e73b2fc19ae21b73a5555c1a726975eabcb/arc-mlir/src/tools/arc.in#L165-L178

segeljakt: Hmm ok I see, it looks like we need this directory structure:
```
arc/
  bin/
    arc-mlir
    arc-lang
    arc
  share/
    arc-runtime/
    stdlib/
      stdlib.arc
      stdlib.mlir
```
I forgot to add the `stdlib`, I'll do it now

frej: > Oh, it looks like you fixed it already

Look at that! My cache is lossy :)

.github/continuous-integration.sh
frej: That sounds much more reasonable,

frej (approved):

PR #415, opened 2022-04-26
Make CI script work with github runners

segeljakt:


frej (changes requested):
`run-build` sets `A2M_CCACHE="1"`, it shouldn't do that when we're running without the `PERSIST_DIR` as it will probably write files to a sibdir of the user's home directory.

.github/continuous-integration.sh
```diff
@@ -56,7 +60,11 @@ function check-ccache {
 }
 
 function run-build {
-    A2M_CCACHE="1" BUILD_FLAVOUR="Release" A2M_ASSERTS="1" run-step ./build
+    if [[ -z "${PERSIST_DIR}" ]]; then
+        BUILD_FLAVOUR="Release" A2M_ASSERTS="1" run-step ./build
```
frej, line 64: Duplicating command lines are not a good thing, we should do it like we do it in arc, i.e.

```shell
A2M_CMAKE_DEFS=('BUILD_FLAVOUR="Release"' 'A2M_ASSERTS="1"')

if [[ ! -z "${PERSIST_DIR}" ]]; then
  A2M_CMAKE_DEFS+=('A2M_CCACHE="1"')
fi
...
"${A2M_CMAKE_DEFS[@]}" run-step ./build
```

segeljakt: I updated the script now 👍

.github/continuous-integration.sh
```diff
@@ -56,7 +60,13 @@ function check-ccache {
...
 
 function run-build {
-    A2M_CCACHE="1" BUILD_FLAVOUR="Release" A2M_ASSERTS="1" run-step ./build
+    A2M_CMAKE_DEFS=('BUILD_FLAVOUR="Release"' 'A2M_ASSERTS="1"')
+
+    if [[ ! -z "${PERSIST_DIR}" ]]; then
+        A2M_CMAKE_DEFS+=('A2M_CCACHE="1"')
+    fi
+
+    "${A2M_CMAKE_DEFS[@]}" run-step ./build
```
frej, line 69: Sorry for that suggestion, we have to do `env "${A2M_CMAKE_DEFS[@]}" run-step ./build`for the environment variables to be set.

frej (approved):
If you fix my broken suggestion, the test will hopefully pass.

frej:
```
env: 'run-step': No such file or directory
```
Ah, run-step is a bash function, we have to `export -f` it.

So add `export -f run-step` after the definition.

frej:
Try this
```bash
env "${A2M_CMAKE_DEFS[@]}" bash -c run-step ./build
```

PR #416, opened 2022-05-02
LLVM submodule update

frej:
Changes:

  * MLIR now attaches InferTypeOpInterface on
    SameOperandsAndResultType operations when possible, so the
    implementation of SameOperandsAndResultType operations requires the
    inclusion of the InferTypeOpInterface.h header.


segeljakt (approved):
LGTM

PR #417, opened 2022-05-09
 LLVM submodule update

frej:
Changes:

 * Upstream has removed the special case for parsing `FuncOp` which
   was put in place when moving `FuncOp` to the `func` dialect. We now
   have to use `func.func` in our MLIR.


segeljakt (approved):
LGTM

PR #422, opened 2022-05-23
LLVM submodule update

frej:
Changes:
 * Avoid warning about return value of parseOptionalComma() not
   checked in arc-mlir dialect.


segeljakt (approved):
LGTM

PR #424, opened 2022-06-02
rust-mlir: Handle ADT generics in structs

frej:
Fix a bug in how the Rust dumper handles generic ADT types inside
structs. We have to mangle the type signature for the generic type to
something which is a valid Rust type name when creating the name for
the struct. As the type signature is no longer the same as the Rust
representation of the ADT type, we have also have to modify the MLIR
pretty printer.

### DIFF
--- a/arc-mlir/src/tests/arc-to-rust/adt.mlir
+++ b/arc-mlir/src/tests/arc-to-rust/adt.mlir
@@ -31,4 +31,9 @@ module @arctorustadt {
     -> !arc.generic_adt<"crate::arctorustadt::tests::sharable_Bar::Bar", ui32, !arc.adt<"i32">> {
     return %in : !arc.generic_adt<"crate::arctorustadt::tests::sharable_Bar::Bar", ui32, !arc.adt<"i32">>
   }
+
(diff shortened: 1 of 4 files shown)


arc-mlir/src/tests/arc-to-rust/adt.mlir
```diff
@@ -31,4 +31,9 @@ module @arctorustadt {
     -> !arc.generic_adt<"crate::arctorustadt::tests::sharable_Bar::Bar", ui32, !arc.adt<"i32">> {
     return %in : !arc.generic_adt<"crate::arctorustadt::tests::sharable_Bar::Bar", ui32, !arc.adt<"i32">>
   }
+
+  func.func @ok9(%in : !arc.struct<foo : !arc.generic_adt<"crate::arctorustadt::tests::sharable_Bar::Bar", ui32, !arc.adt<"i32">>>)
+    -> !arc.struct<foo : !arc.generic_adt<"crate::arctorustadt::tests::sharable_Bar::Bar", ui32, !arc.adt<"i32">>> {
+    return %in : !arc.struct<foo : !arc.generic_adt<"crate::arctorustadt::tests::sharable_Bar::Bar", ui32, !arc.adt<"i32">>>
```
segeljakt, line 37: If it helps, I will try to make sure that `::` is not present in ADTs.

segeljakt (approved):
LGTM

PR #433, opened 2022-06-09
rust-mlir: Refactor the Rust pretty printer/dumper

frej:
The previous Rust pretty printer/dumper contained duplicated code and
could not handle nested ADTs properly. It also failed to properly
name-mangle external ADTs.

This is a cleaner re-implementation which changes the code generation
strategy to generate and then use a type alias for all but primitive
types.


segeljakt (approved):
LGTM

PR #440, opened 2022-07-11
LLVM Submodule update

frej:
Changes needed:

  * The parsing of types has changed, the enclosing `"`s are now
    passed on to the TypeStorage. For our dialects it means that we have
    to strip the `"`s when constructing the mangled type for `RustType`.

  * The short type form when the dialect type doesn't contain any
    white-space is no longer done, i.e what previously was printed as
    `!dialect.foo` is no printed as `!dialect<"foo">`. This change
    requires some test-case updates.


segeljakt (approved):
LGTM

PR #445, opened 2022-07-27
Support for the FSM transform

frej:
Tests, which require that pending changes to the runtime system are merged, are in the branch [`frej/fsm-transform-tests`](https://github.com/cda-group/arc/tree/frej/fsm-transform-tests).


segeljakt (approved):
LGTM!

PR #463, opened 2022-12-05
LLVM Submodule update

frej:
Changes:

  * Upstream has now completely removed the support for non-prefixed accessors, so adapt to that. I order to avoid a name clash with accessors for the operands, we also rename the operand names in our dialect specific return instructions.


segeljakt (approved):
LGTM

PR #464, opened 2022-12-12
LLVM Submodule update

frej:
Changes:

  * Switch from the deprecated `llvm::None` to `std::nullopt`.

  * Upstream has switched from required attributes to interface methods for the FunctionOpInterface, so change our implementation accordingly.


segeljakt (approved):
LGTM

PR #465, opened 2022-12-14
rust-mlir: Use func.func and not no longer supported func

frej:
Looks like this was missed when upstream removed `func` from the standard dialect.


segeljakt (approved):
LGTM

PR #475, opened 2023-01-30
LLVM Submodule update

frej:
Changes:

 * Adapt to upstream change "Make PassManager default to op-agnostic" https://reviews.llvm.org/D137731.


segeljakt (approved):
LGTM

PR #480, opened 2023-02-14
arc-mlir: Drop forgotten tensor test cases

frej:
Tensors were removed in 7ceb6fc850db5287e5a60bbfcc88c576b40d5fd1 so no need to keep test cases around.


segeljakt:
LGTM

PR #487, opened 2023-03-22
Copy adidoks website template manually

segeljakt:
It seems like the template files are no longer copied automatically for some reason. Let's see if this fixes it.


frej (approved):
LGTM

PR #494, opened 2023-08-14
Update arc-runner => github-actions in CI

segeljakt:


.github/continuous-integration.sh
```diff
@@ -5,7 +5,7 @@ set -e # Terminates as soon as something fails
 echo "The work dir is ${ARC_MLIR_BUILD}"
 
 export PATH="$ARC_MLIR_BUILD/llvm-build/bin:$PATH"
-export RUSTC_WRAPPER="/home/arc-runner/.cargo/bin/sccache"
+export RUSTC_WRAPPER="/home/github-actions/.cargo/bin/sccache"
```
frej, line 8: Are you sure  /home/github-actions is persisted between invocations of the runner?

segeljakt: It should be ok, this is running in a docker image on my personal computer.

frej: Good

frej (changes requested):
You can't just disable the arc-runtime, it breaks all test cases which produce and compile Rust. I don't care about `arc-lang-to-rust/*` but `arc-to-rust/*` and `rust/*` are important to ensure that pulling in stuff from upstream doesn't break stuff.

segeljakt:
Hmm ok, all the checks went through CI so I got confused.

The runtime has problems upstream because of the Rust nightly version being used. It might be possible to by rolling back to an earlier nightly version. I will try to see if that's the case. Which nightly version are you using right now?

frej:
> Hmm ok, all the checks went through CI so I got confused.

I see one failure. "test / build (pull_request) Failing after 1m"

> Which nightly version are you using right now?

I'm on 1.71.1, although I haven zapped ~/.cargo lately, so maybe I'm saved by having old cached stuff.

segeljakt:
Oh, I got very confused. I was checking the actions https://github.com/cda-group/arc/actions page and thought the first run represented all the checks 😅

segeljakt:
I was able to roll back to version 1.70.0 and things seem to work

segeljakt:
The `klas/v1` branch, and future updates, will only rely on the stable branch to avoid these issues

frej (approved):

PR #495, opened 2023-10-04
LLVM Submodule update

frej:
Changes needed:

 * FunctionInterfaces.h has moved to the mlir/Interfaces/ directory.

 * FunctionOpInterface is now a subclass to CallableOpInterface, so getCallableRegion() has to be implemented for RustFuncOp and RustExtFuncOp.

 * The binop folder now supports poison, so we explicitly have to disable the poison support for our integer operations.


segeljakt (approved):
LGTM. I forgot that the runner died because of a power outage so I will restart it

